Issue #21, opened 2026-06-09
Discord session silently parks on Claude folder-trust prompt after CUSTOMER_HOME migration


jacketyjax:
## Summary

Discord session silently parks on Claude Code's first-run folder-trust dialog after the `#6` CUSTOMER_HOME migration, on any install whose CUSTOMER_HOME path has no trust entry in `~/.claude.json`. Bot appears alive (tmux session exists, process is healthy) but never connects to Discord. Watchdog does not detect it.

Reported and root-caused by @tobyzn on his `asimov` install: full writeup at https://gist.github.com/tobyzn/2e70b548e94d497b18f71b7ac6c9f407 - this issue lifts his analysis upstream so the fix lands in chassis for every install going forward.

## Symptom

- Discord bot stops replying. No crash, no error logs.
- `tmux ls` shows the `*-discord` session is alive.
- `scripts/` directory is intact (so this is **not** the 2026-06-02 "re-bootstrap deleted scripts" incident from `#new-jaxity#XXX`).
- `tmux capture-pane -t <session> -p | tail -30` reveals the claude process is parked on:

  ```
  Accessing workspace: /Users/<user>/.behalfbot
  Is this a project you trust?
   > 1. Yes, I trust this folder
     2. No, exit
  ```

## Root cause

1. Claude Code stores folder-trust state **per absolute path** in `~/.claude.json` under `projects["<abs path>"].hasTrustDialogAccepted`.
2. `--dangerously-skip-permissions` does **not** bypass the folder-trust gate. It controls the per-tool permission classifier, which is a separate downstream check that only fires after trust is established.
3. The `#6` CUSTOMER_HOME migration changed the Discord launch cwd from CHASSIS_HOME (`~/behalfbot`, already trusted on pre-migration installs) to CUSTOMER_HOME (`~/.behalfbot`, **never trusted** on any install).
4. Result: every Discord restart on a post-migration install boots Claude into a path with no trust entry, hits the interactive prompt, and waits for keyboard input forever.

### Secondary contradiction in restart template

`chassis/scripts/templates/restart-discord.sh.template` has a comment claiming the working dir is CHASSIS_HOME:

> Working dir is CHASSIS_HOME so claude resolves CLAUDE.md from there.

…but the rendered command actually launches from CUSTOMER_HOME:

```bash
tmux new-session -s "<name>-discord" -d \
[...]

Issue #6, opened 2026-06-02
chore(chassis): customer state out of chassis tree so container is disposable


jacketyjax:
## Problem

The chassis tree at \`~/behalfbot/\` mixes chassis code (disposable, re-pullable) with customer state (must survive reinstall). The survivor dirs (\`briefings/\`, \`data/\`, \`logs/\`, \`memory/\`, \`scheduled-tasks/\`, \`state/\`, \`temp/\`) only survive a reinstall because they're gitignored. Anything customer-side that ISN'T in the gitignore list dies silently on \`rm -rf ~/behalfbot && git clone\`.

This bit Toby on 2026-06-02. He got a fresh chassis install on 2026-06-01 16:08 (open-source migration). His customer-side \`~/behalfbot/scripts/\` directory - which held the restart-asimov-discord.sh and watchdog-asimov-discord.sh that the LaunchAgent invokes - got nuked because it wasn't in the gitignore. The asimov-discord tmux session that was already in memory kept running fine through the night - the symptom only surfaced when Toby shut down his Mac for the night. On the morning reboot, the watchdog LaunchAgent fired exit 127 (\"command not found\") because the restart script was gone. So the SILENT FAILURE window was just this morning, not all night - but the silent failure mode is real: presence shows online (stale Discord state from the dead tmux), watchdog keeps firing exit 127, no error escalation.

This will hit every installer who re-bootstraps or pulls a fresh chassis. Silent failure mode: the LaunchAgent looks alive (it's loaded and firing), the Discord presence looks online (stale from prior tmux session), but no messages get processed. Daily-briefing dispatcher also fails because the morning-briefing prompt references scripts that were deleted.

## Proposed design — hard separation

Customer state lives OUTSIDE the chassis tree:

\`\`\`
~/behalfbot/                  # chassis clone, FULLY disposable
~/.behalfbot/                 # customer state, NEVER touched by reinstall
  scripts/                    # restart/watchdog/bot-specific
  state/
  memory/
  briefings/
  logs/
  HEARTBEATS.md               # per-customer heartbeat config
  chassis.config.yaml.local   # per-customer config overrides
\`\`\`

- Bind-mount the customer dirs into the container where chassis code expects them.
- LaunchAgent plists reference \`~/.behalfbot/scripts/restart-\${BOT_NAME}-discord.sh\`.
- Bootstrap on first install: create \`~/.behalfbot/\` if missing, scaffold from chassis-carried templates, NEVER overwrite if present.
[...]

PR #22, opened 2026-06-09
fix(bootstrap): pre-seed claude folder-trust for CUSTOMER_HOME (#21)

jacketyjax:
## Summary

Fix for #21: bootstrap now pre-seeds Claude Code's per-directory folder-trust state for `CUSTOMER_HOME` in `~/.claude.json`, so Discord launches never park on the first-run trust prompt.

- `--dangerously-skip-permissions` does **not** bypass the trust gate. Trust is stored per absolute path in `~/.claude.json` under `projects["<abs path>"].hasTrustDialogAccepted`.
- The `#6` CUSTOMER_HOME migration moved the Discord launch cwd to `${CUSTOMER_HOME}` (e.g. `~/.behalfbot`), a path no install had ever trusted - so every restart since #6 has been one trust-prompt away from a silent outage. Toby's `asimov` hit it; full reproduction and manual recovery procedure in his gist: https://gist.github.com/tobyzn/2e70b548e94d497b18f71b7ac6c9f407
- `bootstrap-customer-scripts.sh` now writes:

  ```json
  {
    "projects": {
      "<CUSTOMER_HOME>": {
        "hasTrustDialogAccepted": true,
        "hasCompletedProjectOnboarding": true
      }
    }
  }
  ```

  ...idempotently merged into `~/.claude.json` after the customer-side scripts are rendered.

## Edge cases handled

| Case | Behavior |
|---|---|
| `~/.claude.json` exists, valid JSON, no trust entry for CUSTOMER_HOME | Merge entry, preserve all other top-level keys |
| `~/.claude.json` exists, trust entry already present and correct | No-op (logs "already set") |
| `~/.claude.json` does not exist | Create minimal config with just the projects entry |
| `~/.claude.json` exists but is invalid JSON | Refuse, exit 1, prompt operator to fix or delete |
| `~/.claude.json` exists but `projects` is not an object | Refuse, exit 1 (rather than clobber) |
| `python3` not on PATH | Warn, skip pre-seed, continue bootstrap |
| `--dry-run` | Logs intent, writes nothing |

Write is atomic (`tempfile.mkstemp` + `os.replace`) so a partial write can't corrupt `~/.claude.json` if the process is interrupted mid-edit.

## Test

Manually verified all four primary cases against an isolated `$HOME` (valid-JSON merge, idempotent rerun, invalid JSON, no file). Output matches expectations; preexisting `~/.claude.json` keys are preserved across the merge.

## Scope

- This PR is the **causal fix** for #21 (eliminate the failure mode).
- A separate follow-up PR will harden the watchdog to detect "tmux alive but bot parked on a prompt" - covers any future variant of this class of bug. Splitting so each can land and roll back independently.
- The `restart-discord.sh.template` cwd/comment contradiction noted in Toby's gist is already resolved on `main` (comment now accurately says "Working dir is CUSTOMER_HOME"), no fix needed there.

## Credit

Bug reported, root-caused, and manually recovered by @tobyzn on his `asimov` install. Gist: https://gist.github.com/tobyzn/2e70b548e94d497b18f71b7ac6c9f407.

Refs #21.

## Test plan

- [ ] @tobyzn tests on `asimov`: revert the manual `~/.claude.json` edit, re-run `bash chassis/scripts/bootstrap-customer-scripts.sh`, confirm `~/.claude.json` is repopulated and the next `bash scripts/restart-asimov-discord.sh` boots straight to "Listening for channel messages" with no prompt.
- [ ] Jax re-runs bootstrap on his own box, confirms idempotent no-op and no corruption.
- [ ] Verify `--dry-run` does not touch `~/.claude.json`.

### DIFF
--- a/chassis/scripts/bootstrap-customer-scripts.sh
+++ b/chassis/scripts/bootstrap-customer-scripts.sh
@@ -148,6 +148,95 @@ render_template "$TEMPLATES_DIR/restart-discord.sh.template" \
 render_template "$TEMPLATES_DIR/watchdog-discord.sh.template" \
     "$OUT_SCRIPTS/watchdog-${BOT_NAME}-discord.sh"
 
+# Pre-seed Claude Code's per-directory folder-trust state for CUSTOMER_HOME.
+#
+# Claude Code stores folder-trust per absolute path in ~/.claude.json under
+# projects["<abs path>"].hasTrustDialogAccepted. --dangerously-skip-permissions
+# does NOT bypass this gate; it controls the per-tool permission classifier,
+# which only runs after trust is established. If CUSTOMER_HOME is not already
+# trusted, claude boots into an interactive trust prompt and parks forever -
+# the Discord tmux session stays alive (so the watchdog session-existence check
+# passes) but the bot never connects. Reported by Toby on asimov via #21.
+#
+# This step is idempotent: re-running merges fields rather than clobbering
+# unrelated state.
+preseed_claude_trust() {
+    local target_dir="$1"
+    local claude_config="$HOME/.claude.json"
+
+    if [[ "$DRY_RUN" == "true" ]]; then
+        echo "[dry-run] would pre-seed claude folder-trust for $target_dir in $claude_config"
+        return 0
+    fi
+
+    if ! command -v python3 >/dev/null 2>&1; then
+        echo "  WARN: python3 not on PATH; skipping claude-trust pre-seed for $target_dir" >&2
+        echo "        First Discord launch will hit the trust prompt and park." >&2
+        return 0
+    fi
+
+    if ! python3 - "$claude_config" "$target_dir" <<'PYEOF'
+import json
+import os
+import sys
+import tempfile
+
+config_path, target_dir = sys.argv[1], sys.argv[2]
+
+if os.path.exists(config_path):
+    try:
+        with open(config_path) as f:
+            data = json.load(f)
+    except json.JSONDecodeError as e:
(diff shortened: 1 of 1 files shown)


tobyzn:
**Tested on asimov (real install, macOS) — PASS ✅**

Reverted to the broken state (deleted `projects["/Users/tobyscregg/.behalfbot"]` from `~/.claude.json`), then ran `bootstrap-customer-scripts.sh`:

- Pre-seeds `hasTrustDialogAccepted` + `hasCompletedProjectOnboarding` via atomic tempfile write ✅
- Idempotent — re-run prints `already set (no-op)` ✅
- `--dry-run` previews without writing ✅
- Restart afterward reaches `Listening for channel messages` with no trust prompt ✅

This does exactly what it claims. Two notes for when this + #23 land together:

**1. Branch topology.** As pushed, this PR and #23 are siblings off `c8cec74` (#20) — `#23` does **not** contain this commit (`git merge-base --is-ancestor <this> pr-23` → false). Testing #23 alone would lack the bootstrap pre-seed. I tested the **merged combination** for the watchdog work; results are on #23.

**2. The recovery path doesn't re-seed trust (cross-ref #23).** The pre-seed lives only here in `bootstrap-customer-scripts.sh`. The restart template touched by this PR does **not** pre-seed trust, and the #23 watchdog recovers by calling the **restart script**, not bootstrap. So if trust is ever wiped (the exact root-cause condition), watchdog → restart → re-park → loop, with nothing on that path re-seeding. Details + empirical repro on #23. Worth factoring the pre-seed into a shared helper that both bootstrap and the restart script call.

PR #58, opened 2026-07-09
feat(second-brain): Stage 1 - secondbrain MCP server, second_brain.mode, list_recent

jacketyjax:
Stage 1 of the second-brain cutover, per the ratified plan (one chassis-owned `secondbrain` MCP server over `get_adapter()`, `second_brain.mode: direct | adapter` defaulting to `direct`, native backend MCP not registered in adapter mode). Later stages (Pacman queue to Postgres, daily-log migration, prompt rewrites) are deliberately NOT in this PR.

**Inert for existing installs.** A config with no `mode` key renders `.mcp.json` exactly as today - verified by test (`test_missing_mode_means_direct`) and by a CLI dry-run of `hydrate-mcp-json.py` against a legacy fixture config (output servers: `context7, github, memory, playwright, siyuan` - identical to main).

## What changed

- `chassis.config.yaml`: `second_brain.mode` key, documented default `direct`.
- `chassis/second_brain/factory.py`: `get_mode()` - missing key means `direct`, invalid values raise.
- `chassis/second_brain/mcp_server.py` (new): the `secondbrain` MCP server. Six tools mirroring the `NotesAdapter` protocol: `create_doc`, `append_to_doc`, `read_doc`, `search`, `list_recent`, `get_deeplink`. Resolves its backend via `get_adapter()` at startup, so identical tool names work against SiYuan, Notion, or Obsidian. **There was no MCP server precedent in this repo, so it uses the official Python MCP SDK** (`mcp==1.28.1`, added to `requirements.txt`) over stdio, as the task allows.
- `chassis/second_brain/base.py`: `list_recent(since, until, min_content_len, limit)` added to the `NotesAdapter` protocol, implemented on all three adapters (SiYuan SQL, Obsidian mtime scan, Notion `last_edited_time`). Callers are NOT migrated in this stage.
- `chassis/.mcp.json.template`: new `secondbrain` entry gated on `mode == 'adapter'`; `siyuan` and `notion` entries additionally gated on `mode != 'adapter'` so the native backend server is omitted entirely in adapter mode (tool availability is the guardrail).
- `chassis/scripts/hydrate-mcp-json.py`: `_enable_when` grammar grows `!=` and `&&`. Missing-key semantics are deliberate: `==` on a missing path is False, `!=` on a missing path is True - that is what keeps configs that predate the `mode` key on the direct path.
- `docs/second-brain-adapters.md`, `docs/mcp-setup.md`: mode semantics, the server, and the `list_recent` divergence table.
- `.github/workflows/python-tests.yml` (new): first pytest workflow in the repo (previously NO CI ran the Python suites), scoped to the second-brain paths.

## (backend, mode) matrix

| backend | direct | adapter |
|---|---|---|
| siyuan | `siyuan` server (as today) | `secondbrain` only |
| notion | `notion` server (as today) | `secondbrain` only |
| obsidian | **no second-brain MCP server - effectively invalid** | `secondbrain` only |

`(obsidian, direct)` gets no entry because no suitable Obsidian MCP server exists to register: the community options require the Obsidian desktop app plus its Local REST API plugin, which a headless container install does not run. I did not fabricate one. Obsidian installs are adapter-mode-only, stated in the template divider, `chassis.config.yaml` comments, and both docs. This matches the status quo (the template had no obsidian option before this PR either).

## `list_recent` parity, honestly

Per Sean's instruction, all three backends implement it with the closest honest approximation; divergences are in the protocol docstring and `docs/second-brain-adapters.md`:

- **SiYuan**: SQL over `blocks.updated` (second granularity, kernel-local clock). Cleanest signal.
- **Obsidian**: filesystem mtime scan. Noisier than SiYuan - a git pull or iCloud resync produces false "activity". `min_content_len` is approximated by file size in bytes.
- **Notion**: `/search` sorted by `last_edited_time` descending with client-side windowing, because the endpoint has no timestamp filter. Minute granularity (Notion truncates seconds). Only pages shared with the integration are visible. `min_content_len > 0` costs one extra API call per candidate page. Scan capped at 500 pages per call.

## Where the plan did not survive contact with the code

1. **The plan's "SiYuan: the existing SQL, verbatim" is wrong.** `daily-log-gather.py:326-336` filters doc rows with `LENGTH(content) > 200`, but `blocks.content` for `type='d'` rows holds the TITLE, not the body. Verified against Sean's live kernel: `MAX(LENGTH(content))` over all 283 `type='d'` rows is 81, so that filter can never match and the production `siyuan_activity` gather is very likely returning an empty list every night. The adapter's `list_recent` instead filters on `SUM(LENGTH(content))` over child blocks via a correlated subquery (also verified against the live kernel, read-only). The `daily-log-gather.py` fix itself is a later stage per scope; flagging it now so it does not get ported verbatim.
2. **`bootstrap-mcp-config.sh` has a pre-existing gap the plan does not mention**: it strips `_enable_when` keys without evaluating them (`chassis/scripts/bootstrap-mcp-config.sh:154`), so its sed-based render emits EVERY server regardless of backend or mode. The bootstrap path that matters (`bootstrap.sh:314` calling `hydrate-mcp-json.py`) is mode-aware after this PR; the recovery script was already backend-unaware before it. I left it untouched to keep Stage 1 inert - it should be taught to delegate to `hydrate-mcp-json.py` in the stage that makes the mode flip operational.
3. **Smoke-test rework**: the plan's Stage 1 bullet includes reworking `smoke-test.sh:209-225` per backend, but the task instructions explicitly scope it out of this PR. Task instructions win; not touched.
4. **Server path in the template**: the entry uses `${CHASSIS_ROOT:-/app/chassis}/second_brain/mcp_server.py`. `CHASSIS_ROOT` is baked into the container image (`Dockerfile:74`, chassis tree copied to `/app/chassis` at `Dockerfile:317`) and Claude Code expands `${VAR:-default}` in `.mcp.json` at load time. Host-side (non-container) layouts need the vendored path instead - stated in the entry's `_install_note`. Mild uncertainty flag: I verified the expansion syntax against Claude Code's documented `.mcp.json` env-expansion behavior, not against a live container; the e2e test drives the server by absolute path, which is layout-independent.

## Test plan

- [x] Suite before: 30 passed. Suite after: 68 passed (30 existing untouched + 38 new). `python3 -m pytest chassis/second_brain/tests/ -q`
- [x] `list_recent` unit tests on all three adapters against throwaway fixtures (temp vault with controlled mtimes for Obsidian; stubbed `_post` / `_request` transports for SiYuan and Notion - window filtering, half-open interval, ordering, `min_content_len`, limit, pagination)
- [x] `.mcp.json` render matrix: all six (backend, mode) combinations plus the missing-mode compatibility case, driven through the REAL template and REAL hydrator (`test_mcp_json_render.py`)
- [x] End-to-end: spawned `mcp_server.py` as a subprocess over stdio against a throwaway Obsidian vault, exactly as Claude Code launches it. Observed: initialize handshake returns `serverInfo.name == "secondbrain"`; `tools/list` returns exactly the six tools; `create_doc` wrote `Briefings/manual-drill.md` into the vault and returned `{id, deeplink}`; `read_doc` round-tripped the body; `append_to_doc` content visible in the next read; `search` found seeded content; `list_recent` returned the just-created doc in a 2-hour window and nothing in a future window; `get_deeplink` returned `obsidian://open?vault=...`; a malformed `since` datetime surfaced as a tool error (`isError: true`) with the server still answering afterwards (`test_mcp_server.py` runs this same drill in CI)
- [x] Legacy-config CLI dry-run: `hydrate-mcp-json.py --dry-run` with a mode-less siyuan config renders the same server set as main
- [x] MCP e2e tests skip cleanly (not fail) on an interpreter without the `mcp` package
- [x] Secret scan of the diff: clean. No em dashes in added lines. `chassis/VERSION` untouched per policy.

Stage 2 and beyond (Pacman queue to Postgres, `daily-log-gather.py` adapter branch, prompt rewrites, smoke-test backend dispatch) land as separate PRs per the staged plan.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,6 +52,19 @@ VAULTWARDEN_DOMAIN=http://localhost:8222
 # Token: SiYuan Settings > About > API token.
 # SIYUAN_URL=http://host.docker.internal:6806
 # SIYUAN_TOKEN=
+#
+# Deeplink prefix. A block id is appended to this verbatim, so it MUST end with
+# whatever separator the target expects (a trailing `=` or `/`).
+#
+# The chassis default is `siyuan://blocks/`, which opens the SiYuan DESKTOP APP.
+# It does NOT open on a phone. For links that are clickable on mobile, point this
+# at your SiYuan web UI instead - the shape is:
+#
+#   https://<your-siyuan-host>:6806/stage/build/desktop/?id=
+#
+# Substitute your own reachable hostname (public DNS, Tailnet, reverse proxy).
+# Set it here, in ONE place: the host moves, the code should not.
+# SIYUAN_DEEPLINK_BASE=
 
 # Notion. Internal integration token from notion.so/my-integrations. The target
 # pages must be explicitly shared with the integration.
--- a/.env.example
+++ b/.env.example
@@ -40,3 +40,22 @@ VAULTWARDEN_ADMIN_TOKEN=CHANGEME_run_openssl_rand_base64_32
 # public-facing URL.
 VAULTWARDEN_HOST_PORT=8222
 VAULTWARDEN_DOMAIN=http://localhost:8222
+
+# --- Second brain -----------------------------------------------------------
+# Credentials for the backend named in chassis.config.yaml `second_brain.backend`.
+# ONE source for both modes: the native MCP server (mode: direct) and the
+# chassis `secondbrain` adapter server (mode: adapter) read these same vars.
+# Set only the pair matching your backend; leave the rest unset.
+#
+# SiYuan. From inside the chassis container the kernel lives on the HOST, so
+# host.docker.internal - NOT 127.0.0.1, which resolves to the container itself.
+# Token: SiYuan Settings > About > API token.
+# SIYUAN_URL=http://host.docker.internal:6806
+# SIYUAN_TOKEN=
+
+# Notion. Internal integration token from notion.so/my-integrations. The target
+# pages must be explicitly shared with the integration.
+# NOTION_API_TOKEN=
+
+# Obsidian needs no credential - set second_brain.obsidian.vault_path in
+# chassis.config.yaml instead.
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -1,0 +1,41 @@
+name: Python tests
+
+# First pytest workflow in the repo (added with the second-brain Stage 1 PR -
+# before this, no CI ran the Python suites at all). Scoped to the paths the
+# suite actually covers so unrelated PRs don't pay for it.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'chassis/second_brain/**'
+      - 'chassis/scripts/hydrate-mcp-json.py'
+      - 'chassis/.mcp.json.template'
+      - 'chassis.config.yaml'
+      - 'requirements.txt'
+      - '.github/workflows/python-tests.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'chassis/second_brain/**'
+      - 'chassis/scripts/hydrate-mcp-json.py'
+      - 'chassis/.mcp.json.template'
+      - 'chassis.config.yaml'
+      - 'requirements.txt'
+      - '.github/workflows/python-tests.yml'
+  workflow_dispatch:
+
+jobs:
+  pytest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      # Install only what the suite needs, not the full chassis image deps
+      # (psycopg/pillow/boto3 are runtime-only and slow the job for nothing).
+      # mcp is pinned to the same version as requirements.txt so the e2e
+      # server test exercises what the image ships.
+      - run: pip install pytest pyyaml mcp==1.28.1
+      - run: python3 -m pytest chassis/second_brain/tests/ -v
--- a/chassis.config.yaml
+++ b/chassis.config.yaml
@@ -125,8 +125,50 @@ second_brain:
   organization_pattern: para               # Tiago Forte: Projects / Areas / Resources / Archives
   databases:
     primary_crm: TBD_at_install             # Notion database id
-    notes_root: TBD_at_install              # Page id for free-form prose writes
+    notes_root: TBD_at_install              # Page id for free-form prose writes. Doubles as the
+                                            # default write target for the adapter (SiYuan
+                                            # notebook_id / Notion notes_root) unless the
+                                            # per-backend block below overrides it.
   scaffold_para_if_missing: true            # If their Notion isn't PARA-shaped, scaffold it
+
+  # ---------------------------------------------------------------------------
+  # Per-backend blocks - OPTIONAL OVERRIDES ONLY. Leave them commented out.
+  #
+  # CREDENTIALS COME FROM .env, NOT FROM HERE. Adapter mode (`mode: adapter`)
+  # and direct mode read the SAME vars, so a secret is never duplicated:
+  #
+  #   siyuan   -> SIYUAN_URL, SIYUAN_TOKEN
+  #   notion   -> NOTION_API_TOKEN
+  #   obsidian -> no credential; vault_path below is required (it is a path,
+  #               not a secret)
+  #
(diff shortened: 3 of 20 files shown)


jacketyjax:
## Review: two blocking bugs, both fixed on this branch

Reviewed by driving the adapter against a **live SiYuan kernel**, not the test suite. That mattered: CI was green and the PR was mergeable, but every SiYuan test stubs the transport, so the kernel never got a vote on anything this PR emits.

### 1. Adapter mode was dead on arrival (fixed in 928c657)

Direct mode and adapter mode read SiYuan credentials from **disjoint sources**:

- **Direct** (works today): the `siyuan` entry in `.mcp.json.template` gets `SIYUAN_URL`/`SIYUAN_TOKEN` from placeholders that bootstrap substitutes out of `.env`.
- **Adapter** (this PR): `mcp_server.py` resolves `get_adapter()`, which reads `second_brain.siyuan.{base_url,token,notebook_id}` from `chassis.config.yaml` - **a sub-block that exists in no shipped template and in no real install.**

So `get_adapter()` silently constructed `SiYuanNotes(base_url='http://127.0.0.1:6806', token='', notebook_id='')`. Against a live kernel an empty token returns `{"code":-1,"msg":"Auth failed [session]"}`. `main()` resolves the adapter eagerly, so flipping `mode: adapter` would have taken SiYuan completely dark. `127.0.0.1` is also wrong from inside the chassis container, which reaches the kernel at `host.docker.internal:6806`.

Fix: the SiYuan branch of the factory now falls back to the same `SIYUAN_URL`/`SIYUAN_TOKEN` env vars direct mode already uses, and `notebook_id` falls back to the existing canonical `second_brain.databases.notes_root`. An unresolvable token now raises at startup instead of producing an adapter that 401s on every call.

### 2. A SQL `ESCAPE` clause silently zeroed out all search (introduced, then removed, in 8083620)

Hardening `search()` against LIKE wildcards looked correct and was not. Measured against the live kernel:

| statement | result |
|---|---|
| `content LIKE '%Vibecode%'` | 397 rows |
| `content LIKE '%Vibecode%' ESCAPE '\\'` | `data: null` |
| `... ESCAPE '!'` / `... ESCAPE '#'` | `data: null` |

**SiYuan rejects the `ESCAPE` clause outright, whatever the escape character.** Do not re-add it.

### 3. The bug class underneath (the important one)

SiYuan signals a refused query with a **success-shaped** `{"code": 0, "msg": "", "data": null}`. Every SQL call site did `rows = result if isinstance(result, list) else []`, which turns a **rejected query into an empty result set, with no error**. That is why the ESCAPE regression was invisible. A genuinely empty result comes back as `[]`, never `null`, so the distinction is safe to rely on.

All four SQL call sites now go through one `_query_sql()` that raises `SiYuanError` on a null or non-list payload. Non-SQL endpoints that legitimately return null (e.g. `appendBlock`) are untouched.

### Not a bug: `append_block`

I went in expecting a known silent no-op on doc-type parents and believed I had reproduced it. I had not. The write persists correctly; **SiYuan's SQL index is eventually consistent** and lags writes by seconds. `exportMdContent` (which reads the file) showed the content present the whole time.

Real consequence, now in the divergence docs: `search()` and `list_recent()` read that index, so neither can see a doc written moments earlier by `create_doc`/`append_to_doc`. Obsidian (filesystem) and Notion (API) have no equivalent lag.

### Verification

Live kernel, real config, after the fixes: `search("Vibecode")` returns hits again (was 0), `list_recent` pulled real daily logs / the weekly rollup / the CBT-I doc, `read_doc` round-tripped, and a null SQL payload raises. Suite: 84 passing. Both new test groups were shown red on the pre-fix code before going green.

### Still required before any cutover (install-side, not merge blockers)

1. `SIYUAN_TOKEN` is not in the `.env` of the reference install - it only lives inline in `.mcp.json`. Adapter mode needs it in `.env`.
2. No `deeplink_template` is set, so deeplinks emit as `siyuan://blocks/<id>` rather than the phone-clickable `https://s.grid7.com/?id=<id>`. That would break links in briefings.

Stage 1 remains inert for existing installs: `mode` absent still means `direct`, and the rendered `.mcp.json` is byte-identical to main across all backend/mode combinations.

*Follow-ups worth issues, deliberately not in this PR: Notion adapter mode has the same missing-config-block problem and still fails soft rather than loud; `notes_root` ships as the literal `TBD_at_install`, which is non-empty and so passes validation only to fail later at the kernel.*

Issue #53, opened 2026-07-07
Architecture: centralize shared plugins/skills across behalfbot + BehalfBot-open


jacketyjax:
## TL;DR recommendation

Extract shared plugins, skills, agents, guardrail policy, and the trigger/parser core into a third repo, **`scrollinondubs/behalfbot-plugins`**, and have both chassis vendor it with **git subtree** at a `plugins/` prefix - the exact pattern installs already use to vendor the chassis itself. Vendor the WHOLE repo into every chassis; the interview selects what to **activate**, not what to vendor. Sean's third-repo lean is validated - the copy-paste alternative is already producing measurable drift after one day.

### behalfbot (Claude Code chassis)

```
behalfbot/
├── plugins/                      # 7 product plugins
│   ├── angel-protocol/  bfl/  dating/  loom-vision/
│   ├── remarkable/  restaurant-booking/  whatsapp/
│   │   ├── openclaw.plugin.json      # harness-NEUTRAL manifest (contracts.tools/skills/hooks/triggers, configSchema)
│   │   ├── .claude-plugin/plugin.json # Claude Code compat shim (4 of 7 have it)
│   │   ├── skills/*.md               # flat .md skill format
│   │   ├── scripts/  scheduled-tasks/  db/migrations/  triggers/
├── chassis/skills/               # chassis-core skills: chassis-update.md, pacman.md
├── chassis/scripts/              # dispatcher, parsers/, merge-plugin-triggers.sh, gather-*
├── chassis/.claude/hooks/        # guardrails.sh, principal-policy-hook.sh (bash enforcement)
└── chassis.config.yaml           # modules: toggles decide which plugins activate at bootstrap step 9
```

### BehalfBot-open (opencode/Venice chassis)

```
BehalfBot-open/
├── .opencode/skills/             # brand-voice, content-research, devops, sdr-outreach (SKILL.md dir format)
├── .opencode/agents/             # software-engineering, startup-advisor, security-advisor
├── .opencode/plugins/guardrails.ts   # TS reimplementation of the bash hook policy
├── scripts/                      # dispatch-trigger.sh, parsers/, heartbeat-dispatcher.sh (forked from chassis)
└── docs/PORTING.md               # 6-phase porting plan; explicitly notes everything should land upstream
```

### Shared vs chassis-specific vs drifted

| Asset | behalfbot | BehalfBot-open | Verdict |
|---|---|---|---|
[...]

PR #75, opened 2026-07-18
feat(plugins): runtime-pull plugin machinery - Phase 0 consumption (behalfbot#53)

jacketyjax:
Phase 0 consumption machinery for the runtime-pull plugin decision on #53 (Option B - install-time fetch). Nothing activates until Sean creates `scrollinondubs/behalfbot-plugins`, tags it, and lands a pin: `chassis/PLUGINS_PIN` ships UNPINNED, so every install keeps running the image-baked `/app/plugins` tree byte-identically.

## What it does

- **Fetch**: `chassis/scripts/fetch-plugins.sh` reads `chassis/PLUGINS_PIN` (`<tag> <40-hex-sha>`, sibling of `chassis/VERSION`), resolves the tag via `git ls-remote`, downloads the tarball at the pinned SHA, sanity-checks the tree against `registry.json`, atomically swaps it into `$CUSTOMER_HOME/vendored-plugins/`, and writes `$CUSTOMER_HOME/plugins.lock` (repo, tag, commit, per-plugin `{version, tag, sha}`). `--freeze` / `--unfreeze` for air-gapped installs. Boot-time fetch is best-effort - a network failure never takes the bot down; the explicit `update-plugins` entrypoint mode propagates failures.
- **Activation**: `bootstrap.sh` step 9 `activate_plugins()` is no longer a log-only stub. It calls `chassis/scripts/activate-plugins.sh`, which discovers enabled plugins across the layers, runs each plugin's `setup.sh`, writes `$CUSTOMER_HOME/chassis-env.sh` from manifest `contracts.env`, merges `contracts.mcpServers` into `.mcp.json` under `"_managed_by": "behalfbot-plugin:<name>"` markers (re-runs replace managed entries, manual entries untouched), and re-merges triggers.
- **Layering** (first hit wins): `$CUSTOMER_HOME/plugins-local/` (customer-private plugins, never published) > `$CUSTOMER_HOME/plugins/` (legacy customer-local, e.g. midnight-oil) > `$CUSTOMER_HOME/vendored-plugins/` (fetched) > `/app/plugins` (image-baked offline fallback during migration).

## The 4 adversarial-review fixes applied (from the #53 review)

1. **Fetch destination is NOT `$CUSTOMER_HOME/plugins`** (finding 1). That dir holds live customer-local plugins and is never written by the fetcher. Fetch lands in `$CUSTOMER_HOME/vendored-plugins/` - on the existing customer bind mount, writable by the runtime uid, survives container recreate. `merge-plugin-triggers.sh`'s hardcoded `chassis_home/"plugins"` lookup is replaced with the layered search path as an in-scope code change, not an assumed-true one.
2. **SHA is the gate, not the tag** (finding 3, fleet-RCE). `PLUGINS_PIN` records `<tag> <sha>`; the fetcher resolves the tag remotely and REFUSES to fetch (exit 3, previous tree untouched) when the resolved SHA differs from the pin. A force-moved tag never auto-refetches. A bare tag with no SHA is rejected as malformed. Verified live against a real public tag with a mismatched pin.
3. **Mixed image/clone states degrade gracefully** (finding 2). `entrypoint.sh` and `bootstrap.sh` execute from the image, so old-image + new-clone keeps booting byte-identically (no fetch step exists there), and new-image + stale-clone finds no pin/activator and WARN-skips onto the baked tree. The activation and fetch MEAT lives in chassis scripts that ride the clone overlay, so post-release fixes there do not need another image release.
4. **Real `activate_plugins()`** (original review's weakest-assumption finding). Enabled-gating via a shared parser (`lib/enabled-plugins.py`, extracted so `merge-plugin-triggers.sh` and the activator cannot drift), `setup.sh` execution with WARN-not-fatal failures, env contract emission, managed MCP registration, trigger merge across layers. Functionally tested: layering precedence, idempotent re-runs, manual `.mcp.json` entries preserved.

## Migration note - NOT zero-touch

Because `entrypoint.sh` and `bootstrap.sh` are image-baked, activating this machinery requires an **image release + container recreate**. A chassis clone refresh alone changes nothing at boot. Until that release, this PR is inert everywhere.

## Sean-gated items (none of them in this PR)

- Creating `scrollinondubs/behalfbot-plugins` (public; only public-safe plugins - private ones stay install-side in `plugins-local/`)
- First tag + first pin line in `PLUGINS_PIN`
- chassis VERSION bump (untouched here, still 0.1.1)
- Image release / recreate

Do not merge without Sean's ratification.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -526,23 +526,48 @@ preflight_bot_identity() {
 activate_plugins() {
     step "9/14" "Activate enabled plugins"
 
-    log "TODO: for each plugin in plugins/ directory:"
-    log "  - Read its openclaw.plugin.json"
-    log "  - Check chassis.config.yaml.modules.<plugin>.enabled"
-    log "  - If enabled: source plugin's activation hook + export its env vars"
-    log "  - Write chassis-env.sh that the launchd/systemd unit sources"
-    log ""
-    log "Per-plugin chassis-env.sh exports (V1-known):"
-    log "  whatsapp: CHASSIS_WHATSAPP_SAFE=\$CHASSIS_PLUGINS_ROOT/whatsapp/scripts/wacli-safe.sh"
-    log "  bfl:      BFL_ARCHIVE_DIR=<from config>"
-    log "  dating:   SOCIAL_CHANNEL_ID=<from config>"
-    log "  etc."
-    log ""
-    log "Plugin trigger merge (per chassis/scripts/merge-plugin-triggers.sh):"
-    log "  - Reads contracts.triggers from each enabled plugin's openclaw.plugin.json"
-    log "  - Writes merged registry to \$CUSTOMER_HOME/triggers.yaml"
-    log "  - Read by chassis/scripts/dispatch-trigger.sh on every inbound message"
-    log "  - Re-run any time a plugin is enabled/disabled or its manifest changes"
+    # Real implementation lives in chassis/scripts/activate-plugins.sh
+    # (behalfbot#53 Phase 0 - replaces the former log-only stub). The meat is
+    # kept in a chassis script, not inline here, because bootstrap.sh is
+    # image-baked while chassis scripts ride the clone overlay: fixes to
+    # activation logic ship via clone refresh without an image release.
+    #
+    # It discovers plugins across the layered roots (plugins-local >
+    # customer plugins > vendored-plugins fetched from behalfbot-plugins >
+    # image-baked /app/plugins fallback), gates on chassis.config.yaml
+    # modules.<name>.enabled, runs each plugin's setup.sh, writes
+    # $CUSTOMER_HOME/chassis-env.sh from contracts.env, merges
+    # contracts.mcpServers into .mcp.json (managed markers), and re-merges
+    # plugin triggers via merge-plugin-triggers.sh.
+    local activator=""
+    local candidate
+    for candidate in \
+        "$CHASSIS_HOME/chassis/chassis/scripts/activate-plugins.sh" \
+        "$CHASSIS_HOME/chassis/scripts/activate-plugins.sh" \
+        "${CHASSIS_ROOT:-/app/chassis}/scripts/activate-plugins.sh"; do
+        if [[ -f "$candidate" ]]; then
+            activator="$candidate"
+            break
+        fi
+    done
+
+    if [[ -z "$activator" ]]; then
+        # Mixed-state migration window (new image, stale clone): behave like
+        # the old stub - log and continue, baked plugins still drive.
+        log "  WARN: activate-plugins.sh not found in clone or image - skipping"
+        log "  plugin activation (refresh the chassis clone to enable it)"
+        return 0
+    fi
+
+    if [[ "$DRY_RUN" == "true" ]]; then
+        log "  [dry-run] would run: $activator"
+        return 0
+    fi
+
+    if ! CHASSIS_HOME="$CHASSIS_HOME" CUSTOMER_HOME="${CUSTOMER_HOME:-$CHASSIS_HOME}" \
+        bash "$activator" 2>&1 | tee -a "$TRANSCRIPT"; then
+        log "  WARN: plugin activation reported errors (non-fatal, see above)"
+    fi
 }
 
 # ============================================================
--- a/chassis/PLUGINS_PIN
+++ b/chassis/PLUGINS_PIN
@@ -1,0 +1,15 @@
+# behalfbot-plugins runtime-pull pin (behalfbot#53 Phase 0)
+#
+# Format: single non-comment line "<tag> <40-hex-commit-sha>"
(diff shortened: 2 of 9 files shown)


jacketyjax:
## Adversarial + security review (Fable)

Independent verification of PR #75 + the staged plugin repo at behalfbot-plugins-staging. I did not trust the build's claims - I read the diff, scanned the staging tree, and probed the live behalfbot container.

**Verdict: GREEN** (ship after Sean's normal review). Secret scan: **CLEAN**.

### 1. Secret scan of the staging repo (going public) - CLEAN

- No .env, no private keys, no tokens, no credentials, no .git history. Grep hits for "token"/"key" are benign JS identifiers inside the transcript parser in process-loom.sh.
- Contents are exactly: LICENSE, README, docs/MANIFEST.md, registry.json, tools/fetch-plugins.sh, loom-vision/. Nothing else.
- ZERO sensitive plugins and ZERO references to angel-protocol, bodyguard, burglar-alarm, or dating anywhere in the staging tree. Also no leaked host paths, channel IDs, Tailscale IPs, or installer-specific facts.
- Note (pre-existing, NOT introduced by this PR): the public chassis repo itself already carries genericized plugins/angel-protocol and plugins/dating template trees in-tree. They are templatized (no personal data in what I sampled), but Phase 1/2 should move that class of plugin to plugins-local guidance and drop them from the public tree rather than let them ride along.

### 2. Fetch destination - VERIFIED non-colliding

- Destination is $CUSTOMER_HOME/vendored-plugins, not $CUSTOMER_HOME/plugins. Probed the live container: /app/customer is a rw virtiofs bind mount, uid 501 (installer) write-tested OK, survives recreate. vendored-plugins/ does not exist yet - unoccupied.
- /app/customer/plugins currently holds midnight-oil and restaurant-booking; nothing in the diff ever writes there. The atomic-swap logic only touches vendored-plugins and its dot-prefixed staging/previous siblings. No clobber path exists.

### 3. Pinning - tag + SHA, actually enforced

- PLUGINS_PIN requires "<tag> <40-hex-sha>"; a bare tag fails the regex and exits 3 with no fetch. Tag is resolved via git ls-remote (peeled ref preferred); resolved SHA != pinned SHA exits 3, previous tree untouched. Download itself is BY SHA (codeload tarball at the SHA, git-clone fallback checks out the exact SHA), so even a TOCTOU tag move cannot change what lands on disk. A moved upstream tag cannot trigger auto-refetch or RCE.
- The pin file preference order favors the clone overlay, and the clone is mounted read-only in the container (verified: ro virtiofs), so an in-container prompt-injected agent cannot tamper with the pin. Pin bumps require a reviewed chassis PR. Freeze flag honored for air-gapped installs.
- Hardening notes (not blockers, same trust domain as existing agent code-exec): (a) the on-disk vendored tree is never re-hashed against the lock SHA, so a tamper on the rw mount persists and the lock suppresses refetch; (b) chassis-env.sh writes values with naive double quotes - a manifest env value containing "$( )" would execute when sourced. Use printf %q or shlex.quote in a follow-up.

### 4. plugins-local layering - genuine

Search order plugins-local > plugins > vendored-plugins > baked is implemented in both activate-plugins.sh and merge-plugin-triggers.sh (first manifest hit wins), and entrypoint creates plugins-local/. A private plugin on the reference install loads without ever existing in the public repo. Confirmed the staging repo carries only loom-vision.

### 5. Migration honesty - honest

docs/runtime-pull-plugins.md says it outright: entrypoint.sh and bootstrap.sh are image-baked, so activation requires an image release plus container recreate. Verified against the live container: the running image's entrypoint has zero fetch_plugins/update-plugins references and no fetch/activate scripts under /app/chassis/scripts. Both mixed states (new image + stale clone, refreshed clone + old image) degrade to today's behavior with a WARN. No zero-touch claim anywhere.

### 6. activate_plugins() correctness - real, not a stub

It discovers across the four layers, gates on modules.<name>.enabled via the shared enabled-plugins.py, runs setup.sh (WARN on failure), generates chassis-env.sh, merges contracts.mcpServers into .mcp.json with _managed_by markers (stale managed entries removed, manual entries never touched, 0600 atomic write), and re-runs the trigger merge. Two correctness notes: (a) dispatcher-mode boot fetches but does not re-activate, so a pin bump absorbed by a dispatcher restart updates plugin code while chassis-env.sh/.mcp.json/triggers.yaml stay stale until the next bootstrap - new triggers or env vars from an updated plugin are silently missing; consider running activation after a successful fetch, or documenting "bootstrap after pin bump" as mandatory. (b) enabled-plugins.py only recognizes the inline-bool form at exactly indent 2 - fine for the current config shape, brittle if it ever changes.

### The one blocker before this ships value

Nothing here activates until Sean cuts a chassis image release and recreates the container (entrypoint is image-baked), then creates scrollinondubs/behalfbot-plugins from the staging tree, tags it, and lands the tag+SHA pin in the same PR as the VERSION bump. Until that sequenced, Sean-gated release happens, this PR is safe, dormant machinery - which is exactly what Phase 0 promised.

PR #130, opened 2026-07-31
fix(daily-log-gather): make merged/opened/closed_unmerged day filters exclusive on the upper bound

### LINKED ISSUE #126: daily-log-gather.py counts today's merged PRs as yesterday's work
## What happens

`chassis/scripts/daily-log-gather.py` filters merged PRs with an inclusive comparison on both ends of the window:

```python
since_str = since.date().isoformat()   # yesterday
until_str = until.date().isoformat()   # today
...
if merged_at and since_str <= merged_at[:10] <= until_str:
    merged.append(item)
```

`until_str` is today. So every PR merged today passes the filter and lands in a payload whose `date` field says yesterday. The `--search updated:{since}..{until}` argument on the `gh pr list` call has the same inclusive shape, so nothing upstream of the comparison narrows it either.

`metrics.prs_merged` is derived from the same buckets, so the count inherits the error.

## Evidence

Run on Sean's install, 2026-07-30, `DAILY_LOG_GH_USER=jacketyjax`:

```
$ chassis/scripts/daily-log-gather.py --now 2026-07-19T02:00:00Z
-> 27 merged PRs reported for date 2026-07-18

$ gh search prs --merged-at 2026-07-18 --author jacketyjax --json url | jq length
8
$ gh search prs --merged-at 2026-07-19 --author jacketyjax --json url | jq length
19
```

27 = 8 + 19. The payload is a two-day union labelled as one day.

Live run the same evening reported 21 merged for `date: 2026-07-29`, against a true count of 10 for that date and 11 for 2026-07-30.

## Why it has not been noticed

The heartbeat runs at 02:00, so only PRs merged between 00:00 and 02:00 UTC leak in. On this install that is currently zero: of 174 PRs merged 2026-07-16..29, none closed in that band. The bug is invisible in production and loud on any manual or backfill invocation.

It is not zero-impact though. A PR merged at 00:30 UTC is counted in both the previous day's log and the current day's, since the next night's window starts at that same midnight.

Any consumer that runs later than 02:00 is exposed proportionally. A morning-briefing-style caller firing between 06:00 and 11:00 would have picked up 24 of those 174 PRs as the wrong day's work, which is why the install kept a customer-side gather for that path rather than cutting it over (scrollinondubs/new-jaxity#309).

## Fix

Make the upper bound exclusive:

```python
if merged_at and merged_at[:10] == since_str:
```

or keep the range form and compare against `until` exclusive rather than `until_str` inclusive. Same treatment for `opened` and `closed_unmerged`, which use the identical pattern.

Worth a test in `chassis/scripts/test-daily-log-gather.sh` that asserts the window is one calendar day wide. The current suite runs every scenario with credentials unset, so no scenario exercises a populated `prs_by_repo` at all.

Found while wiring Sean's install off its local fork and onto this script (scrollinondubs/new-jaxity#309).

jacketyjax:
## Summary
- `since_str <= x <= until_str` treated PRs/issues from *today* as belonging to *yesterday*'s report, since the payload's `date` field is `since_str` (yesterday). A PR merged at 00:30 UTC landed in both the previous day's log and the current one.
- Extracted the per-PR classification into `classify_pr_for_day()` and compared only against `since_str`, matching the single-day semantics the report actually claims.
- Applied identically to `merged`, `opened` (OPEN + createdAt), and `closed_unmerged` (CLOSED + closedAt, not merged).

## Test plan
- [x] `chassis/scripts/test-daily-log-gather.sh` - added scenario 7, a direct unit test of `classify_pr_for_day` against fixtures spanning the since/until boundary (mirrors the bug report's 2026-07-18/19 evidence). All 7 scenarios pass locally.
- [x] `python3 -m py_compile chassis/scripts/daily-log-gather.py`

Closes #126

### DIFF
--- a/chassis/scripts/daily-log-gather.py
+++ b/chassis/scripts/daily-log-gather.py
@@ -177,6 +177,23 @@ def run_text(argv: list[str], *, verbose: bool, timeout: int = 60) -> str | None
 # --- GitHub surface -------------------------------------------------------
 
 
+def classify_pr_for_day(pr: dict, day_str: str) -> str | None:
+    """Bucket a PR dict into "merged" / "opened" / "closed_unmerged" for the
+    single calendar day `day_str` (YYYY-MM-DD), or None if it belongs to a
+    different day. Each PR maps to at most one bucket for a given day."""
+    state = (pr.get("state") or "").upper()
+    merged_at = pr.get("mergedAt")
+    closed_at = pr.get("closedAt")
+    created_at = pr.get("createdAt")
+    if merged_at and merged_at[:10] == day_str:
+        return "merged"
+    if state == "OPEN" and created_at and created_at[:10] == day_str:
+        return "opened"
+    if state == "CLOSED" and closed_at and not merged_at and closed_at[:10] == day_str:
+        return "closed_unmerged"
+    return None
+
+
 def gather_github(gh_user: str, since: datetime, until: datetime, *, verbose: bool
                   ) -> tuple[dict[str, dict], list[dict], list[str]]:
     """Return (prs_by_repo, open_issues_awaiting_input, warnings)."""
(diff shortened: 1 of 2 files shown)


jacketyjax:
Superseded by #132 - same diff, commit now carries the DCO `Signed-off-by` trailer. Could not amend in place because force-push is blocked by `.claude/hooks/guardrails.sh:34`.

PR #131, opened 2026-07-31
fix(validate-heartbeats): strip env-prefixed gather args before path resolution

### LINKED ISSUE #127: validate-heartbeats.sh flags any gather line with an inline env prefix as a missing file
## What happens

`chassis/scripts/validate-heartbeats.sh` strips inline arguments off a `gather:` value by taking the first whitespace token:

```bash
# Strip any inline arguments (e.g. `gather: scripts/foo.sh --json`)
gather_path="${value%% *}"
```

That handles trailing flags but not a leading environment prefix, which is the documented way to parameterise a shared chassis gather per heartbeat. `gather: PG_CONTAINER=jax-pg BACKUP_SUBDIR=jax-pg chassis/chassis/scripts/pg-backup.sh` resolves `gather_path` to `PG_CONTAINER=jax-pg` and errors:

```
ERR  [jax-pg-backup].gather references missing file: `PG_CONTAINER=jax-pg`
     (resolved to `/app/customer/PG_CONTAINER=jax-pg`).
```

Sean's install has three such heartbeats and therefore three permanent false errors, out of four total. The linter's signal-to-noise is now bad enough that a real stale path would be easy to miss in the output.

## Fix

Drop leading `KEY=VALUE` tokens before taking the path:

```bash
gather_path="$value"
while [[ "$gather_path" == *=* && "${gather_path%%[ =]*}" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]]; do
    gather_path="${gather_path#* }"
done
gather_path="${gather_path%% *}"
```

Also worth skipping a leading `env` token while in there.

Found while wiring Sean's install onto the chassis daily-log gather (scrollinondubs/new-jaxity#309), which adds a fourth heartbeat with this shape.

jacketyjax:
## Summary
- `gather_path="${value%% *}"` only stripped trailing inline args, not a leading `KEY=VALUE ...` env prefix - the documented way to parameterise a shared chassis gather per heartbeat.
- `gather: PG_CONTAINER=jax-pg BACKUP_SUBDIR=jax-pg chassis/chassis/scripts/pg-backup.sh` resolved `gather_path` to `PG_CONTAINER=jax-pg` and permanently flagged the heartbeat as a missing file. Sean's install had three such heartbeats, all permanently false-erroring.
- Now drops a leading `env` token and any leading `KEY=VALUE` tokens before taking the path, then strips trailing inline args as before.

## Test plan
- [x] Added `chassis/scripts/test-validate-heartbeats.sh` - 6 scenarios: plain path (baseline), one env prefix, two env prefixes (the exact shape from the bug report), leading `env` token, env-prefix + trailing args together, and a genuinely-missing file (guards against false negatives from the new stripping logic). All pass.

Closes #127

### DIFF
--- a/chassis/scripts/test-validate-heartbeats.sh
+++ b/chassis/scripts/test-validate-heartbeats.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+# test-validate-heartbeats.sh - Regression test for validate-heartbeats.sh's
+# gather-path resolution.
+#
+# The bug this locks down
+# ========================
+# `gather:` lines can carry a leading `KEY=VALUE ...` env prefix (the
+# documented way to parameterise a shared chassis gather per heartbeat, e.g.
+# `gather: PG_CONTAINER=jax-pg BACKUP_SUBDIR=jax-pg chassis/chassis/scripts/pg-backup.sh`).
+# The old `gather_path="${value%% *}"` took the first whitespace token, which
+# resolved to `PG_CONTAINER=jax-pg` instead of the script path, and every such
+# heartbeat permanently failed the missing-file check.
+#
+# No docker daemon, no network. Everything runs against a scratch HEARTBEATS.md
+# + scratch chassis tree under a temp CHASSIS_HOME.
+#
+# Exit codes:
+#   0 - all scenarios passed
+#   1 - one or more scenarios failed
+#   2 - test harness itself broke
+
+set -uo pipefail
+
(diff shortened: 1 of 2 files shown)


jacketyjax:
Superseded by #133 - same diff, commit now carries the DCO `Signed-off-by` trailer. Could not amend in place because force-push is blocked by `.claude/hooks/guardrails.sh:34`.

PR #136, opened 2026-08-07
feat(gmail-monitor): read attachments, and stop returning empty bodies for HTML-only mail

jacketyjax:
Two holes in the gmail-monitor read surface, both invisible from outside the server.

## What was wrong

**HTML-only mail returned nothing.** `_extract_plaintext` only ever walked `text/plain` parts. Most machine-sent mail is HTML-only, so `get_thread` returned `body: ""` with no signal that anything had been missed - the caller could not distinguish an empty message from an unreadable one.

Concretely, this morning on Sean's install: a Telnyx single-use sign-in link came back empty. The token had to be reconstructed by hand from the UUIDv7 timestamp against the `Date` header just to get into the account.

**Attachments were unreachable.** No tool fetched them, and `get_thread` did not even list them, so there was no way to tell a message carrying a PDF from one that wasn't.

## What changed

- `get_thread` reports `body_mime` and falls back to rendering `text/html` when there is no `text/plain` part. **The renderer keeps link targets** - a sign-in link or document URL living only in an `href` is exactly the content worth having, and a naive tag strip throws it away. Link text is emitted as `label (url)`, not `label <url>`, because the generic tag strip would eat the angle-bracketed form. Found that one in test.
- `get_thread` lists every attachment inline: `filename`, `mime_type`, `size_bytes`, `attachment_id`.
- **New `get_attachment`** writes bytes to disk and returns a path ready to `Read`, rather than inlining base64 - a PDF is not useful as base64 in a transcript. Selects by `attachment_id`, by `filename` (exact then substring), or implicitly when the message has exactly one.
- **New `list_attachments`** for when you hold a `message_id` and nothing else.

Still no send tool. Draft-only by construction is unchanged.

## Safety

Filenames from email are attacker-controlled. Path separators, NUL and control bytes are replaced and the name is basenamed, so `../../etc/passwd` lands as `passwd`. It is a **blocklist, not an allowlist**, so `Apólice.pdf` survives as itself instead of arriving as `Apo_lice.pdf`.

Downloads are capped by `GMAIL_MONITOR_MAX_ATTACHMENT_BYTES` (default 25 MB) and land in `GMAIL_MONITOR_ATTACHMENT_DIR` (default `<install>/data/gmail-attachments`).

## Verified against live Gmail, not mocks

| Check | Result |
|---|---|
| Download a 171,297 byte insurance PDF | sha256 **identical** to the same file fetched independently via Discord |
| Select by filename substring | correct file, 171,297 bytes |
| Implicit single-attachment select | correct file |
| Unknown `attachment_id` | returns the available list rather than throwing |
| Telnyx message that returned an empty body | now `body_mime: text/html`, full sign-in URL with UUID intact |
| Filename traversal / unicode / empty / 200-char | all handled |
| `py_compile`, template JSON parse, stdio handshake | clean; `tools/list` advertises seven |

## Deploy note

This is chassis code, so an install picks it up on the next subtree pull plus an MCP restart. Nothing changes for an install that never enables `gmail_monitor`.


jacketyjax:
Superseded by #137: same diff, same single commit, now carrying a `Signed-off-by` line so the DCO check passes. Reopened on a new branch rather than force-pushing this one.

PR #143, opened 2026-08-09
fix(audit): Gap 3 resolves the memory graph, not the config shape (#142)

### LINKED ISSUE #142: bootstrap-audit Gap 3 always fails on the current .mcp.json.template (checks memory.env.MEMORY_FILE_PATH, template sets none)
## Summary

`scripts/bootstrap-audit.sh` Gap 3 reads `.mcpServers.memory.env.MEMORY_FILE_PATH` from the customer's `.mcp.json`. The current `.mcp.json.template` deliberately does not set an `env` block for the memory server: it resolves the path from the launch cwd inside a `sh -c` wrapper instead. So a correctly-hydrated install fails Gap 3.

Worse than a cosmetic false alarm: Gap 3 is the check that is supposed to catch an unreachable memory graph, and it cannot pass on any install running the current template. It has therefore never been able to catch the failure it exists for.

## Reproduction

Verified 2026-08-09 on Sean's install, inside the chassis container:

```
$ bash /app/chassis/scripts/bootstrap-audit.sh
...
Gap 3: memory MCP wired with writable path
  ✗ mcpServers.memory.env.MEMORY_FILE_PATH not set
```

with a `.mcp.json` memory block that matches the template and demonstrably works:

```json
"memory": {
  "command": "sh",
  "args": [
    "-c",
    "export MEMORY_FILE_PATH=\"$(pwd)/memory/memory.jsonl\"; exec npx -y @modelcontextprotocol/server-memory"
  ]
}
```

```
$ python3 -c "import json; d=json.load(open('/app/chassis/.mcp.json.template')); print('env' in d['mcpServers']['memory'])"
False
```

Memory MCP itself is working on that install: an entity written from a container session was read back from a separate container session and from a host session, all against the one bind-mounted `memory/memory.jsonl`.

## Root cause

`scripts/bootstrap-audit.sh` around lines 222-242 assumes the pre-`sh -c` shape of the template. The template moved to cwd-resolution to fix the real bug, which is that the host and the container see the same bind-mounted directory at different absolute paths (`/Users/<user>/.behalfbot` vs `/app/customer`), so a single baked absolute path is valid in only one namespace. The audit was not updated alongside it.

The stale expansion block is a second symptom of the same drift:

```bash
expanded="${expanded//\$\{CHASSIS_HOME\}/${CHASSIS_HOME:-${HOME}/behalfbot}}"
```

It only substitutes the bare `${CHASSIS_HOME}` token, not the `${CHASSIS_HOME:-default}` form, and the template no longer emits either.

## Proposed fix

Stop inferring the path from config shape and check the thing the gap is actually about: whether the memory graph file is reachable and writable from the launch cwd.

1. Accept both shapes. If `.mcpServers.memory.env.MEMORY_FILE_PATH` is set, keep the existing expansion path for backward compatibility with legacy renders.
2. If it is absent, fall back to `$CUSTOMER_HOME/memory/memory.jsonl`, which is what the `sh -c` wrapper resolves to, and run the same writability check against that.
3. Fail only when neither resolves to a writable path.
4. Keep failing loudly on a baked absolute path that does not exist in the current namespace, since that is the original bug and legacy installs can still carry it.

Optional and stronger: replace the static inspection with a live check that spawns the configured server and asserts the graph file exists and is writable after startup. That would have caught the original failure on any install regardless of config shape.

## Impact

Every install on the current template reports a false Gap 3 failure, and no install has a working automated check for an unreachable memory graph. On Sean's install this class of failure went undetected for at least the window between 2026-07-24 and 2026-08-09, during which container-side sessions read an empty graph and wrote to a path that did not exist.

jacketyjax:
Fixes #142.

## What was wrong

`bootstrap-audit.sh` Gap 3 read `.mcpServers.memory.env.MEMORY_FILE_PATH` out of the customer's `.mcp.json`. The current `.mcp.json.template` deliberately sets no `env` block - it exports `MEMORY_FILE_PATH="$(pwd)/memory/memory.jsonl"` inside an `sh -c` wrapper, because the host and the chassis container see the same bind-mounted directory at different absolute paths (`/Users/<user>/.behalfbot` vs `/app/customer`) and a single baked absolute path is valid in only one namespace.

So Gap 3 failed on every install running the current template, and the check that exists to catch an unreachable knowledge graph had never been able to pass, let alone catch anything.

Second symptom of the same drift, same function: the expansion block substituted only the bare `${CHASSIS_HOME}` token, never the `${CHASSIS_HOME:-default}` form.

## What changed

`chassis/scripts/bootstrap-audit.sh`

- Gap 3 resolves the graph path instead of asserting a config shape: legacy `env.MEMORY_FILE_PATH` when present, otherwise the cwd-resolved `$CUSTOMER_HOME/memory/memory.jsonl` that the `sh -c` wrapper produces. A relative path resolves against the customer root, which is the launch cwd on both the host and inside the container.
- A missing graph directory is now a **failure**, not a warning. That is the original amnesiac-bot symptom - an absolute path baked in one namespace and read in the other - not a nit.
- New `expand_home_tokens` handles `${VAR}` and `${VAR:-default}`, scoped to `CHASSIS_HOME` and `CUSTOMER_HOME` only. A set env var wins, then the inline default, then the same fallback bootstrap assumes. Deliberately not a generic expansion: the config is data, and evaluating whatever a hand-edited `.mcp.json` contains is a different bug.
- The pass/fail line now names which shape it resolved from, so a legacy install is distinguishable from a template one in the transcript.

`chassis/scripts/test-bootstrap-audit-memory.sh` (new, wired into `shell-tests.yml`)

Nine cases, five pass-expected and four fail-expected, so the fix cannot degrade into a check that passes everything: both config shapes on a reachable graph, `${CUSTOMER_HOME}`, `${CHASSIS_HOME:-default}` and bare `${CHASSIS_HOME}` forms, then missing graph dir, non-writable graph dir, foreign-namespace absolute path, and an absent memory block. No docker, no network - temp dirs plus jq. The non-writable case self-skips under uid 0 since root ignores mode bits.

Also: `bootstrap-audit-prompt.md` gap-3 description updated to describe the two shapes, and a `CHANGELOG.md` Unreleased entry.

## Why not the live-spawn check

The issue names a stronger option: spawn the configured server and assert the graph file exists and is writable after startup. Skipped, because it would not actually be stronger here. A spawn run by the audit uses the **audit's** cwd, not Claude Code's runtime cwd, so it cannot distinguish "template shape, working" from "template shape, broken" any better than the static resolution against `$CUSTOMER_HOME`. What it would add is an `npx` download and a network dependency in a bootstrap audit, plus - since `server-memory` creates the graph lazily on first write - either a startup assertion that proves nothing or an MCP handshake and a write into the customer's real knowledge graph. The path resolution here is the part that was wrong, and it is now checked directly.

Same reasoning against regex-extracting the export line out of the wrapper's `args`: it re-implements shell expansion and re-couples the audit to a config shape, which is what caused this bug.

## Verification

Run against Sean's live install inside the chassis container, on a demonstrably working memory MCP (the graph is 36 lines and read/written across host and container sessions).

**Before** - `docker exec -w /app/customer behalfbot bash /app/chassis/scripts/bootstrap-audit.sh`

```
Gap 3: memory MCP wired with writable path
  X mcpServers.memory.env.MEMORY_FILE_PATH not set

Summary: 2 passed, 2 warned, 1 failed
```

**After** - same install, patched script copied in with `docker cp` to `/tmp` (nothing under `/app/chassis` touched)

```
Gap 3: memory MCP wired with writable path
  OK memory MCP wired; graph writable: /app/customer/memory/memory.jsonl (cwd-resolved, no env block)

Summary: 3 passed, 2 warned, 0 failed
```

The real `memory.jsonl` was `touch`ed only - still 35175 bytes, 36 lines, unchanged content.

**Negative cases, same container, temp customer homes** - a check that passes everything is worse than the bug:

```
### NEGATIVE 1: host-baked absolute path, read inside the container
  X memory graph dir does not exist in this namespace: /Users/jax/.behalfbot/memory (env.MEMORY_FILE_PATH)

### NEGATIVE 2: template shape, resolved dir is non-writable
  X memory graph not writable: /tmp/tmp.mFspEyNrgm/n2/memory/memory.jsonl (cwd-resolved, no env block)

### NEGATIVE 3: template shape, memory/ dir absent entirely
  X memory graph dir does not exist in this namespace: /tmp/tmp.mFspEyNrgm/n3/memory (cwd-resolved, no env block)

### CONTROL: template shape, memory/ present and writable
  OK memory MCP wired; graph writable: /tmp/tmp.mFspEyNrgm/n4/memory/memory.jsonl (cwd-resolved, no env block)
```

**New suite** - `bash chassis/scripts/test-bootstrap-audit-memory.sh`, run on the macOS host and inside the container, identical output both times:

```
test-bootstrap-audit-memory
  ok   current template shape, memory/ writable (Gap 3 pass)
  ok   legacy env, absolute writable path (Gap 3 pass)
  ok   legacy env, ${CUSTOMER_HOME} token (Gap 3 pass)
  ok   legacy env, ${CHASSIS_HOME:-default} form (Gap 3 pass)
  ok   legacy env, bare ${CHASSIS_HOME}, var set (Gap 3 pass)
  ok   current template shape, memory/ MISSING (Gap 3 fail)
  ok   legacy env, path from another namespace (Gap 3 fail)
  ok   memory/ not writable (Gap 3 fail)
  ok   mcpServers.memory absent (Gap 3 fail)

9 passed, 0 failed
```

**Mutation check** - the suite pointed at the unpatched `bootstrap-audit.sh` from `main`, to prove it can fail and that it fails on exactly the three symptoms in the issue:

```
  FAIL current template shape, memory/ writable: expected Gap 3 to pass, got fail
       |   X mcpServers.memory.env.MEMORY_FILE_PATH not set
  FAIL legacy env, ${CHASSIS_HOME:-default} form: expected Gap 3 to pass, got none
       |   ! parent dir does not exist: ${CHASSIS_HOME:-/var/folders/.../chassis-default}/memory
  FAIL legacy env, path from another namespace: expected Gap 3 to fail, got none
       |   ! parent dir does not exist: /Users/nobody-92003/.behalfbot/memory

  6 passed, 3 failed
```

**Other suites**

- `shellcheck -S error` on both shell files: clean. Full-severity run shows only SC2016 info on the deliberately-single-quoted `${VAR}` literals in the test fixtures.
- `python3 -m pytest chassis/second_brain/tests/ chassis/scripts/tests/`: **320 passed, 7 skipped, 8 subtests passed**. Not triggered by this PR's paths, run anyway.
- `shell-tests.yml` parses and lists the new `bootstrap-audit-memory` job; the new files are in both the `pull_request` and `push` path lists.

## Not verified

- Gap 3 has not been run on an actual legacy install carrying a real baked `env.MEMORY_FILE_PATH`. Sean's install is on the current template, so the legacy path is covered by the test fixtures only.
- The Darwin host-side path was exercised only via the test suite's temp homes, not against a real host-side install run of the full audit (Gap 4 needs LaunchDaemons that only exist on a configured host).

## Test plan

- [x] Gap 3 passes on the live install with the current template
- [x] Gap 3 still fails on a foreign-namespace absolute path
- [x] Gap 3 still fails on a non-writable graph dir
- [x] Gap 3 still fails on a missing graph dir and an absent memory block
- [x] Legacy `env.MEMORY_FILE_PATH` shape still passes
- [x] Suite proven to fail against the unpatched script
- [x] shellcheck clean at `severity: error`
- [x] Existing pytest suites green

### DIFF
--- a/.github/workflows/shell-tests.yml
+++ b/.github/workflows/shell-tests.yml
@@ -28,6 +28,8 @@ on:
       - 'plugins/dating/scripts/gather-dating-swipe.sh'
       - 'plugins/dating/scheduled-tasks/recovery-hooks.d/dating-emulator-recovery.sh'
       - 'plugins/dating/tests/test-emulator-state.sh'
+      - 'chassis/scripts/bootstrap-audit.sh'
+      - 'chassis/scripts/test-bootstrap-audit-memory.sh'
       - '.github/workflows/shell-tests.yml'
   push:
     branches: [main]
@@ -49,6 +51,8 @@ on:
       - 'plugins/dating/scripts/gather-dating-swipe.sh'
       - 'plugins/dating/scheduled-tasks/recovery-hooks.d/dating-emulator-recovery.sh'
       - 'plugins/dating/tests/test-emulator-state.sh'
+      - 'chassis/scripts/bootstrap-audit.sh'
+      - 'chassis/scripts/test-bootstrap-audit-memory.sh'
       - '.github/workflows/shell-tests.yml'
   workflow_dispatch:
 
@@ -99,6 +103,18 @@ jobs:
       # _env.sh resolves docker-exec sessions through the state symlink.
       # No docker daemon, runs against temp dirs.
       - run: bash chassis/scripts/test-chassis-root-resolution.sh
+
+  bootstrap-audit-memory:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # Behavioural coverage for #142: bootstrap-audit Gap 3 read
+      # env.MEMORY_FILE_PATH only, while the current .mcp.json.template sets no
+      # env block, so the check could never pass on a healthy install. The
+      # suite asserts both config shapes pass on a reachable graph AND that
+      # every unreachable / non-writable graph still fails, so the fix cannot
+      # degrade into a check that passes everything.
+      - run: bash chassis/scripts/test-bootstrap-audit-memory.sh
 
   dating-emulator-state:
     runs-on: ubuntu-latest
--- a/chassis/CHANGELOG.md
+++ b/chassis/CHANGELOG.md
@@ -18,6 +18,13 @@ Semver:
 
 ## Unreleased
 
+### Fixed
+
+- **`bootstrap-audit.sh` Gap 3 could never pass, so the memory-graph check had never caught anything (#142).** Gap 3 read `.mcpServers.memory.env.MEMORY_FILE_PATH` out of the customer's `.mcp.json`, but the current `.mcp.json.template` deliberately sets no `env` block - it resolves `$(pwd)/memory/memory.jsonl` inside an `sh -c` wrapper, because the host and the chassis container see the same bind-mounted directory at different absolute paths (`/Users/<user>/.behalfbot` vs `/app/customer`) and a single baked absolute path is therefore valid in only one namespace. Every install on the current template reported a false Gap 3 failure, and the check that exists to catch an unreachable knowledge graph could not pass on any of them. Gap 3 now resolves the graph path from either shape - legacy `env.MEMORY_FILE_PATH` when present, otherwise the cwd-resolved `$CUSTOMER_HOME/memory/memory.jsonl` the wrapper produces - and fails only when the resolved graph is unreachable or non-writable.
+- **A memory graph whose directory does not exist is now a Gap 3 failure rather than a warning.** That is the original amnesiac-bot symptom (an absolute path baked in one namespace and read in the other), not a nit.
+- **`${CHASSIS_HOME:-default}` in a hand-edited `.mcp.json` is expanded.** The old substitution handled only the bare `${CHASSIS_HOME}` token, so the defaulted form fell through as a literal path and failed. Expansion stays scoped to `CHASSIS_HOME` and `CUSTOMER_HOME`: the config is data, not shell to evaluate.
+- New behavioural suite `chassis/scripts/test-bootstrap-audit-memory.sh` (wired into `shell-tests.yml`) covers both config shapes and forces every unreachable-graph case - missing dir, non-writable dir, foreign-namespace absolute path, absent memory block - so the fix cannot degrade into a check that passes everything.
+
 ## v0.4.0 - 2026-07-28
 
 ### Added
--- a/chassis/scheduled-tasks/bootstrap-audit-prompt.md
+++ b/chassis/scheduled-tasks/bootstrap-audit-prompt.md
@@ -10,7 +10,7 @@ The transcript walks five known install gaps:
 
 1. **HEARTBEATS.md backup row** - a regular backup heartbeat is registered. Failure means scheduled backups aren't firing.
 2. **Customer GitHub remote** - `origin` points at the customer's own private repo, not the chassis template. Failure means changes here only exist on this machine.
-3. **Memory MCP** - the knowledge graph MCP is wired into `.mcp.json` and its `MEMORY_FILE_PATH` is writable. Failure means narrative memory across sessions is broken.
+3. **Memory MCP** - the knowledge graph MCP is wired into `.mcp.json` and the graph file it resolves to is writable from the launch cwd. The path comes either from a legacy `env.MEMORY_FILE_PATH` or, on the current template, from `$(pwd)/memory/memory.jsonl`. Failure means narrative memory across sessions is broken - most often an absolute path baked in one namespace (the host customer dir) and read in the other (`/app/customer` inside the container).
 4. **LaunchDaemons** - `com.behalfbot.<bot>-discord-restart` and `-discord-watchdog` are loaded in the system domain. Failure means the Discord bridge has no persistent `claude` process to route into.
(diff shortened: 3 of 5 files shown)


jacketyjax:
Note on the red `shellcheck` check: it is pre-existing and unrelated to this PR.

The only finding is `./chassis/scheduled-tasks/tests/test_state_write_race.sh:1:1: error: ShellCheck only supports sh/bash/dash/ksh/'busybox sh' scripts. [SC1071]` - that file is `#!/bin/zsh`, added in a22251c ("fix(dispatcher): lock the state-file read-modify-write cycle"), and `shell-lint.yml` only excludes `heartbeat-dispatcher.sh` by name. `Shell scripts lint` has been failing on `main` on every push since at least 2026-08-02 (runs 30740467957, 30756430283, 31217157661).

Nothing in this PR is flagged: `shellcheck -S error chassis/scripts/bootstrap-audit.sh chassis/scripts/test-bootstrap-audit-memory.sh` is clean locally, and the CI log lists no other file. Left alone here rather than mixed into a bug fix - the one-line fix is adding `test_state_write_race.sh` to `ignore_names` in `shell-lint.yml`.

PR #170, opened 2026-08-19
fix(bake-env): diff from a scrubbed baseline, not the caller's shell

### LINKED ISSUE #168: bake-env.sh: silently produces nothing when run from a shell that already sourced .env
## Symptom

```
ERR: source produced no new application env vars — check /path/.env format
```

The `.env` format is fine. The message sends you to the wrong place.

## Cause

`bake-env.sh` snapshots `env`, sources `.env` in a subshell, and diffs the two to isolate the vars the `.env` defines. Any shell that already has those vars exported produces an empty diff. That is the normal state of an operator shell whose profile loads the install env, and it is the default state of every agent-driven shell.

## Workaround in use

```
env -i PATH=... HOME=... USER=... CHASSIS_HOME=... CUSTOMER_HOME=... bash chassis/scripts/bake-env.sh
```

## Proposal

Either re-exec under `env -i` internally, or detect the condition (pre-env already contains keys the `.env` defines) and say so: "run me from a clean shell, e.g. env -i ...". Right now the failure blames the wrong artifact, and a re-bake blocked by it also blocks every `compose.sh` command via Guard 5, which is how it turns into a stuck deploy.

Found while re-baking before an image refresh on a live install, 2026-08-19.

jacketyjax:
Closes #168.

## The bug

`bake-env.sh` snapshots `env`, sources `.env` in a subshell, and diffs the two to isolate the vars the `.env` defines. That works exactly once: from a shell that has never loaded the install env.

From any shell that has - an operator whose profile loads it, an agent-driven shell, or simply the second bake in one session - those vars are already on the "before" side. The diff comes back empty and the script exits:

```
ERR: source produced no new application env vars — check /path/.env format
```

The `.env` is fine. The message points at the wrong artifact, and the condition that triggers it is the normal state of a working shell rather than anything unusual.

## Why it is worse than a bad message

`compose.sh` Guard 5 refuses every command, `pull` included, while `.env` is newer than `.env.baked`. So a bake that cannot run is a deploy that cannot run. Hit on a live install on 2026-08-19 during an image refresh: the re-bake was the first step and it blocked everything behind it. The workaround was re-running the whole script under `env -i`.

## The fix

Both dumps now run from the same scrubbed baseline rather than from the caller's environment: `PATH`, `HOME`, `USER`, plus the two path vars the script already requires. The hydration block still finds `bw`/`rbw` and still resolves `~`.

What the caller happens to have exported no longer affects the result, which is also the more correct contract: `.env.baked` should reflect `.env`, not whoever ran the bake.

## Verification

On a live install, from the same warm shell, back to back:

```
# before
$ CHASSIS_HOME=... CUSTOMER_HOME=... DRY_RUN=true bash chassis/scripts/bake-env.sh
ERR: source produced no new application env vars — check /Users/jax/.behalfbot/.env format

# after
$ CHASSIS_HOME=... CUSTOMER_HOME=... DRY_RUN=true bash chassis/scripts/bake-env.sh
[dry-run] would write 136 keys to /Users/jax/.behalfbot/.env.baked (mode 0600)
```

136 keys matches what the same install produces from a cold `env -i` shell, and the toxic-var and host-path strippers still fire identically (`ANTHROPIC_API_KEY` excluded, `CUSTOMER_CLAUDE_DIR` / `MANIFEST` / `REPO` stripped).

`chassis/VERSION` bumped 0.5.0 to 0.5.1.


jacketyjax:
Superseded by #171. The commit was missing its `Signed-off-by` trailer, which is a commit-message fix, so it needed a rewrite rather than an extra commit. Re-pushing the amended branch closed this PR automatically and GitHub refused to reopen it. Same change, same branch, DCO now passing. Signed-off-by is now part of my commit routine for this repo.

Issue #173, opened 2026-08-19
Executor has no build identity, so nobody can tell what it is running


jacketyjax:
Follow-up to the 2026-08-19 post-mortem (scrollinondubs/new-jaxity#533). The executor ran month-old application code for a month and the drift was invisible.

## The gap

The container image bakes a shallow clone of `vibecodelisboa` at build time, so the app code it runs is frozen at the last deploy. Nothing exposes which build that is:

- `/healthz` returns `{"ok":true}`
- `/status` returns the last run's timing and exit code

Neither carries a version, a commit or a build date. Answering "is the executor current" required authenticating to Cloudflare and reading `wrangler deployments list`. A question that costs a credential and a CLI invocation does not get asked casually, which is precisely why a month passed.

Every other component here answers this cheaply. The chassis writes `chassis-root.state.json` with baked and live versions, the smoke test asserts on them, the entrypoint logs them at boot. The executor answers only "I am alive".

## Proposed, in order

**1. Bake the SHA and expose it.** Write the `vibecodelisboa` commit SHA (and the build timestamp) into the image in `build.sh`, return both from `/healthz`. Unauthenticated, since neither is a secret. Then:

```
curl -s https://behalfbot-executor.<sub>.workers.dev/healthz
{"ok":true,"appCommit":"5373e25","builtAt":"2026-08-19T13:27:00Z"}
```

**2. Alarm on drift.** With (1), a heartbeat comparing the deployed SHA against `vibecodelisboa` `main` is a two-line gather. It now has somewhere to shout: `alert_ops` could not deliver anything at all until scrollinondubs/new-jaxity#532 landed today.

**3. Close the deploy gap.** Either a workflow that deploys the executor on merge, or a required check on `vibecodelisboa` PRs touching `src/lib/contribution-ledger/**` and `scripts/behalfbot-*` that says in the PR that an executor deploy is owed. Automatic is better. Visible is the minimum, because right now seven PRs merged with nothing anywhere recording the debt.

**4. Stop saying "the image".** `ghcr.io/scrollinondubs/behalfbot` is the chassis image; the executor container image is a different artifact with a different deploy path. Using the bare word cost a morning of crossed wires on 2026-08-19. Worth a docs pass so no installer inherits the ambiguity.

(1) and (2) are the ones that make this class of failure self-reporting. (3) prevents it. (4) is cheap and stops people talking past each other while they do the other three.

jacketyjax:
[...]

PR #174, opened 2026-08-19
fix(executor): the container image could not boot, and crashed on first request

jacketyjax:
The executor image built from current `main` cannot start.

## Reproduction

Pull or build the image, run it, and it exits immediately:

```
$ docker run --rm <image> 
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/prescan-outcome.mjs'
imported from /app/server.mjs
```

## Cause

The Dockerfile copies exactly one file out of `container/`:

```dockerfile
COPY --chown=executor:executor container/server.mjs /app/server.mjs
```

#166 added `container/prescan-outcome.mjs` and made `server.mjs` import it. The `COPY` was not widened, so the module is simply absent from the image and Node fails at load.

## Why it took a day to find

This is worth recording, because the failure was invisible at exactly the moment it was introduced.

On the reference install the running container instance stayed warm through the deploy and kept serving the **previous** image. The deploy reported success. Three pre-scan runs afterwards completed cleanly - all on the old build, which is separately why a just-merged prompt change appeared not to work.

The break surfaced only on the next genuine cold start, hours later, forced by an unrelated `wrangler secret put`. From that point every container start failed, and the observable symptoms pointed everywhere except here:

- `wrangler containers list` reported the application `active` with a live instance
- `wrangler containers instances` showed the instance `running` with a location and version
- the Worker returned `Failed to start container: The container is not running`
- `wrangler containers ssh` rejected the instance id as invalid

That combination reads as a durable-object or placement problem. Deploying a new Worker version, renaming the durable object to force a fresh one, raising `max_instances`, and finally deleting and recreating the container application all failed, because none of them were the cause. Running the image locally answered it in one command.

## Fix

Copy the directory:

```dockerfile
COPY --chown=executor:executor container/ /app/
```

The next module `server.mjs` imports now ships with it rather than breaking the image on first cold start. The `.test.mjs` file riding along is inert.

## Verification

Rebuilt and deployed to the reference install. The container starts and answers:

```
GET /status -> {"inFlight":null,"lastRun":null}
```

## Worth considering separately

A deploy that produces an unbootable image should not be able to report success. A `docker run --rm <image> node --check /app/server.mjs`, or simply starting the image and probing `:8080`, would have failed this in `build.sh` before anything reached Cloudflare. Related: #173, which asks for build identity on `/healthz` for the same family of reasons - right now nothing distinguishes "the deploy worked" from "the old instance is still answering".

### DIFF
--- a/cloudflare/executor/Dockerfile
+++ b/cloudflare/executor/Dockerfile
@@ -53,7 +53,27 @@ RUN cd /app/vibecodelisboa \
 COPY --chown=executor:executor build-context/vibecodelisboa /app/vibecodelisboa
 
 # In-container HTTP shim (trigger endpoint the Worker forwards to).
-COPY --chown=executor:executor container/server.mjs /app/server.mjs
+#
+# Copy the whole directory, not the single entrypoint file. server.mjs imports
+# ./prescan-outcome.mjs, which arrived with the pre-scan chaining work (#166)
+# while this line still copied exactly one file - so the first image built from
+# that code could not boot at all:
+#
+#   Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/prescan-outcome.mjs'
+#   imported from /app/server.mjs
+#
+# What made it expensive to find: the running instance stayed warm across the
+# deploy and kept serving the OLD image, so the deploy looked clean and three
+# subsequent runs succeeded. The break only surfaced on the next genuine cold
+# start - hours later, triggered by an unrelated secret update, with no obvious
+# link back to the deploy that caused it. Every restart after that failed, and
+# the platform reported the instance as `running` while the Worker could not
+# reach it.
+#
+# Copying the directory means the next module server.mjs imports ships with it
+# rather than breaking the image on first cold start. The .test.mjs file riding
+# along is harmless and keeps the rule simple.
+COPY --chown=executor:executor container/ /app/
 
 USER executor
 WORKDIR /app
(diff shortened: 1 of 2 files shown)


jacketyjax:
Second fault found in the same PR (#166), pushed as 39c58ef.

After fixing the missing module the container booted - and then died on the first `/run`:

```
ReferenceError: runTick is not defined
    at Server.<anonymous> (file:///app/server.mjs:52)
```

#166 deleted `runTick` outright - the 40 lines that spawn the heartbeat - while adding the chaining that was supposed to call it. Nothing replaced it. From the Worker's side this surfaces as `Container suddenly disconnected, try again`, which reads as a platform fault rather than a crash in our own code.

**The two faults masked each other.** The container could not boot, so nothing ever reached the crash; once it booted, the first request killed it. Each looked like a separate infrastructure problem, and an hour went into durable objects, instance caps and placement because that is exactly what the symptoms pointed at.

**So #166 has never run.** Not partially, not degraded. The pre-scan chaining it describes has not executed once, because the file it lives in crashes before reaching it.

This commit restores `runTick` from `58b3cc5^` and wires in the chaining #166 was written for, with its stated constraints intact: only a clean exit chains, a clarification never does, and unparseable output degrades to wait-for-the-next-poke. The six tests that shipped with #166 pass, and the run path is now verified against the built image rather than assumed - `POST /run` returns 202, the tick runs, the container survives.

## The gap that let both through

Nothing in CI ever builds this image and starts it. Both faults are caught by starting the built image, waiting a few seconds, and posting to `/run` on port 8080 - about fifteen seconds of work.

A container whose image cannot boot should not be mergeable, and a deploy that ships one should not report success - especially here, where a warm instance keeps serving the old image and makes the deploy look clean. Worth its own issue alongside #173.

PR #193, opened 2026-09-05
fix(colima): one owner, and a start wrapper that survives an unclean shutdown

jacketyjax:
## What broke

A hard power-off leaves Lima's runtime files behind in `~/.colima/_lima/<instance>/`:

```
ha.pid   vz.pid   ha.sock   ssh.sock
```

After the reboot those pid files point at PIDs the kernel has reissued to unrelated processes. Colima reads them, concludes the VM is alive, and every command after that talks to a socket nobody is listening on. On the reference Mac Mini on 2026-09-05 the stale pid was 789, which by then was `sirittsd`:

```
$ colima list
PROFILE    STATUS     ARCH       CPUS    MEMORY    DISK     RUNTIME
default    Broken     aarch64    4       6GiB      30GiB    docker

$ colima status
colima is not running

$ docker ps
Cannot connect to the Docker daemon at unix:///Users/<user>/.colima/default/docker.sock.
Is the docker daemon running?

$ colima start
FATA[0000] errors inspecting instance: [failed to get Info from ".../ha.sock":
dial unix .../ha.sock: connect: connection refused]

$ tail ~/.colima/default/daemon/daemon.log
waiting 5 secs for VM
waiting 5 secs for VM        # forever
```

Postgres, the note-taking app and the chassis container running the heartbeat dispatcher all stayed down.

**`colima start` can never clear this** - it reads the same stale files on every attempt, so a launchd job that retries it retries the same failure. Two jobs were doing exactly that and neither could recover. `colima stop --force && colima start` is the fix; `stop --force` removes the stale pid and socket files, flips the profile from `Broken` to `Stopped`, and does not touch the VM disk image.

Postmortem: scrollinondubs/new-jaxity#550.

## What changed

**`chassis/scripts/colima-ensure.sh`** - the wrapper. Exits 0 if `docker ps` already answers. Skips the first `colima start` outright when the profile reads `Broken`, because that attempt fails every time and costs a boot minute. Otherwise starts, and on failure runs `colima stop --force` and retries exactly once. Verifies `docker ps` rather than the start exit code, because a start that exits 0 onto a socket still refusing connections is the exact state being fixed. Exits 1 and names what it tried if that did not work; there is no third escalation.

It **never** runs `colima delete` or `colima start --reset`. Both are the commonly suggested fix for a wedged Colima and both destroy the VM disk image, taking every Docker volume with it. Two tests assert it: no code path issues either command, and no line invoking `$COLIMA_BIN` mentions them, so a later edit adding one fails CI rather than review.

It holds a lock so concurrent invocations cannot fight - recovering the reference install by hand meant booting out `postgres-watchdog`, `container-liveness` and `terminal-watchdog` first. Staleness is decided by **age** as well as pid liveness: `kill -0` on a recycled pid succeeds for a completely unrelated process, which is the same defect the whole PR is about, so a pid check alone would deadlock the box permanently on exactly the condition it exists to survive.

**`chassis/launchd/com.behalfbot.colima.plist.template`** - the single owner, now shipped from chassis. Daemon domain per `docs/launchd-domains.md`: no login keychain, no `claude`, no shared tmux server, and it must come up on an unattended reboot with nobody logged in. Rendered and registered only where Colima is actually the Docker runtime, so Linux installs and Macs without Colima are never made to sudo for a plist that would no-op. The label is deliberately not bot-scoped - Colima is a per-machine singleton, so two installs on one Mac must share one owner.

`RunAtLoad` plus a 300s `StartInterval` replaces `KeepAlive { SuccessfulExit: true }`, which respawns on exit **zero** and leaves the job alone when it fails. The one case that needed a retry was the one case launchd never retried; the old job sat parked at exit status 1 while Docker stayed down, and its own comment claimed the opposite.

**`bootstrap-customer-scripts.sh`** renders and activates it, warns about any other launchd job that appears to start or restart Colima, and no longer treats a daemon-domain plist in `/Library/LaunchDaemons/` as stale (that check flagged every entry in `CHASSIS_PLIST_DOMAINS`, so it would have refused to proceed on every correctly configured macOS install the moment a daemon-domain plist was added).

**`docs/colima-recovery.md`** - the error signature for whoever greps their way there next, plus the migration. **`docs/installer-homework.md`** gains the Tailscale per-node key-expiry disable.

## Was there a duplicate colima owner inside chassis?

**No.** `grep -rI colima` over `origin/main` returns nothing outside `.git`. Chassis shipped no colima plist and no script that starts Colima, and the two jobs that raced at boot are both customer-side. Nothing in chassis needed demoting; the chassis-side wait-for-docker pattern in `restart-discord.sh.template` and `watchdog-discord.sh.template` was already correct.

## Required customer-side change (this PR cannot make it)

Silently rewriting an installer's own launchd job is not something the bootstrap should do, so this is documented rather than automated. `bootstrap-customer-scripts.sh` now scans both launchd domains on every plist render, follows one level of indirection into the shell scripts those plists invoke, and warns with the exact remediation.

On the reference install there are two:

1. **`~/Library/LaunchAgents/com.jax.postgres.plist`** runs, inline:

   ```sh
   colima status || colima start --cpu 4 --memory 6 --disk 30
   cd ~/.behalfbot/db/postgres && docker compose up -d
   ```

   Change the prefix to a wait:

   ```sh
   for _ in $(seq 1 60); do docker info >/dev/null 2>&1 && break; sleep 3; done
   cd ~/.behalfbot/db/postgres && docker compose up -d
   ```

2. **`~/Library/LaunchAgents/com.jax.postgres-watchdog.plist`** escalates to `colima restart` from `watchdog-postgres.sh`, but only behind an explicit `POSTGRES_WATCHDOG_ALLOW_COLIMA_RESTART=true` that defaults to off. Leave the default off and it is not an owner in practice. No change needed.

Then re-run activation to replace the old daemon in place (the label is unchanged, so it boots out the old job):

```sh
CUSTOMER_HOME=~/.behalfbot CHASSIS_HOME=~/behalfbot BOT_NAME=<bot> \
  bash chassis/scripts/bootstrap-customer-scripts.sh --activate-plists
```

## Test plan

- [x] `bash chassis/scripts/test-colima-ensure.sh` - **31 passed, 0 failed**. Stubs both `colima` and `docker` and overrides the OS probe, so it needs no VM, no daemon and no Mac.
- [x] **The suite discriminates.** Run against a naive `colima start` stand-in, **14 assertions fail**: the Broken recovery, the stale-socket escalation, the docker-probe escalation, both no-op cases, and the lock stand-down. It catches the defect rather than merely passing.
- [x] Destructive-command guard proven: no scenario issues `colima delete` or `colima start --reset`, and the static check over `$COLIMA_BIN` invocations passes.
- [x] Wrapper run end to end against the **live Colima 0.10.1** on the reference host. Healthy path: parsed the real `colima list --json`, logged `docker already answering (profile 'default' status Running). Nothing to do.`, exited 0, left the profile `Running`, released its lock, all 8 containers untouched. The recovery branches are covered by the stubbed suite only - I did not break a production Colima to exercise them live.
- [x] Full render through `bootstrap-customer-scripts.sh --plists` into a temp `CUSTOMER_HOME`: `com.behalfbot.colima.plist` renders, `plutil -lint` OK, **zero unexpanded `${...}` tokens**, `AbandonProcessGroup` reads `true`.
- [x] Second-owner scan run against the real host: finds `com.jax.postgres` (inline) and `com.jax.postgres-watchdog` (via `watchdog-postgres.sh`), and false-positives on none of the two dozen other plists. The first draft matched on the `StartInterval` key and a prose comment; the tightened version does not.
- [x] Process-group behaviour verified rather than assumed - the reason `AbandonProcessGroup` is insurance and not load-bearing:
      `ps -o pid,pgid,ppid,comm -p $(cat ~/.colima/_lima/colima/ha.pid)` gives `8850 8850 1 limactl`. PGID equals PID and PPID is 1, so Lima's hostagent detaches into its own process group and reparents to init. It was never in the wrapper's group, and booting out the old resident `-f` daemon during migration will not take a running VM down with it.
- [x] `shellcheck -S error` clean on all three shell files; `bash -n` clean.
- [x] Tailscale verify command run on the host before documenting it. The first draft (`grep -A2 '"Self"'`) matches nothing; shipped `tailscale status --json | jq '.Self.KeyExpiry'`, which returns `null` on this host, plus a `python3` fallback.
- [x] `no-customer-state-in-chassis` patterns run locally against the tree: no violations.
- [x] `chassis/VERSION` 0.5.1 -> 0.5.2 with the entry under `## Unreleased`, matching how 0.5.1 was cut (`306dcb1` bumped VERSION and added no version heading; `## Unreleased` is rolled up by a separate `chore(release)` commit, and `_chassis-changelog.sh`'s drift digest needs that heading to keep existing).

## Deliberately left out

- **The 90s docker wait in the discord templates.** A `stop --force` plus `start` cycle can plausibly exceed 30 iterations x 3s, so those jobs may now time out on a recovery boot where they previously only raced a cold start. Real, separable, and out of scope here.
- **`--cpu` vs `--cpus` on the old postgres plist.** `colima start --help` on 0.10.1 documents `-c, --cpus`, so the old plist's `--cpu 4` may have been failing flag parsing outright, which would be a second reason it never recovered anything. I could not probe it without risking a VM create on a production host, so I am asserting nothing about it. The wrapper uses the documented `--cpus` spelling and only ever on a profile that does not exist yet.
- **Sizing policy.** The wrapper never passes resource flags to an existing profile. A recovery run is the worst moment to reconfigure a VM. First-ever creation can set `COLIMA_ENSURE_START_ARGS`.


jacketyjax:
## CI status: green on everything this PR touches

`colima-ensure` passes. The one red check, **`plugin-root-resolution`, is pre-existing on `main`** and unrelated: it is the only failing job in `main`'s own latest `shell-tests` run, and this branch touches no plugin or resolver file (`git diff --name-only origin/main...HEAD | grep -E 'plugin|resolve'` is empty).

## Two bugs CI caught that local testing did not, both worth reading

The suite was written and passing on macOS. Running it on a Linux runner found two real defects, which is the argument for it being there.

**1. `stat -f %m` is not an mtime read on GNU coreutils** (fixed in `d1c58e4`)

`-f` means "filesystem status" there. `%m` is read as another FILE operand, so stat prints a whole filesystem block for the directory on stdout and *then* exits non-zero for the missing `%m`. Chained BSD-then-GNU with `||` inside one command substitution, both commands' stdout concatenates and you get a multi-line block with an epoch glued on the end. That parsed as age 0, so every stale lock read as fresh.

The branch this made unreachable on Linux is the recycled-pid lock breaker - the one that exists because the postmortem's stale `ha.pid` pointed at a live, unrelated `sirittsd`. A lock that can only be broken by a dead pid deadlocks forever on exactly that. Same class of defect as the bug the PR is fixing, one layer down.

Now one substitution per assignment, GNU spelling first, and both the reader and its caller validate that what came back is a number.

**2. The "docker is not installed" scenario was resolving the runner's real Docker** (fixed in `fdd723f`)

Scenarios 8 and 8b set `PATH=/usr/bin:/bin` to arrange an absent binary. A GitHub Actions runner ships a real `/usr/bin/docker` with a live daemon behind it, so 8b resolved it, `docker ps` answered, and the wrapper took its "already healthy" exit instead of the branch under test.

It passed on macOS and in a bare `ubuntu:22.04` container and failed only on CI, which is precisely the failure mode a test asserting "this binary is missing" exists to rule out. Both scenarios now run against a PATH holding symlinks to exactly the four externals `colima-ensure.sh` needs and nothing else. `bash` is one of them, because a `PATH=... bash "$ENSURE"` prefix assignment also governs the lookup of `bash` itself and omitting it exits 127 while proving nothing.

## Re-verification after both fixes

| Environment | Result |
|---|---|
| macOS 15 (real Colima 0.10.1 on the box) | 31/31 |
| `ubuntu:22.04`, bare | 31/31 |
| `ubuntu:22.04` with a working `docker` and `colima` planted on PATH (reproduces the runner) | 31/31 |
| GitHub Actions `ubuntu-latest` | pass |
| Against a naive `colima start` stand-in | **14 assertions fail**, so it still discriminates |

Issue #147, opened 2026-08-09
chassis update-check is blind to code that lands under an unchanged VERSION, and the reference install is in that state now


jacketyjax:
## Summary

Split out of #146 (defect 2). The cooldown fix in #146 makes an unanswered offer come back. It does nothing for the case where the version number stops moving while the code keeps moving, which is the state the reference install is in right now.

`chassis/VERSION` only changes on explicit release commits. Everything merged in between lands under `## Unreleased` in the changelog with VERSION untouched. The update checker compares `local VERSION` against `main`'s VERSION, so once those are equal it emits `{"count": 0, "reason": "up_to_date"}` on every subsequent check, regardless of how much new code has landed on main under that number.

## Observed on the reference install, 2026-08-09

Ground truth, not inference:

- The live chassis tree the dispatcher runs is a clone bind-mounted at `/app/customer/chassis` (`chassis/chassis/scripts/...` per the `HEARTBEATS.md` gather line, confirmed in the 2026-08-03 dispatcher log).
- That clone was pulled to `main` on 2026-08-07, landing at `2829d25` with `chassis/VERSION` = `0.4.0`.
- Upstream `main` VERSION is also `0.4.0`.
- So the checker now reports:

```
$ CHASSIS_HOME=<copy> CUSTOMER_HOME=<copy> bash gather-chassis-update-check.sh
{"count": 0, "reason": "up_to_date"}
```

The install is NOT up to date. #142 and #143 both merged on 2026-08-09 under `## Unreleased`, after the 2026-08-07 pull. #142 is the fix for a `bootstrap-audit.sh` memory-graph check that could never pass. The install does not have it, and under the current checker it will never be told, because the two VERSION strings match and will keep matching until someone cuts a release.

Two further things this instance shows:

- Pulling `main` self-applies unreleased code under a released version number. `state/chassis-update/last-applied.json` on that install still reads `{"from": "0.1.1", "to": "0.2.0"}` while the tree runs 0.4.0-plus-unreleased, because the update arrived by `git pull` rather than through `chassis-update.sh`.
- This is not install-specific. Every install that tracks `main` has the same blind spot, and Ben, Marc and Toby all hit it the moment their VERSION matches upstream's.

## Proposed fix

The script already fetches `CHANGELOG.md`, so the cheapest signal available for free is the content of the `## Unreleased` section. Store a hash of that section in `last-offered.json` alongside the version, and treat a changed hash under an unchanged version as re-offerable.

[...]

PR #194, opened 2026-09-05
feat(credential-expiry): warn before a credential expires, and stop the oauth bridge warning into a void

jacketyjax:
Four fixes from the 2026-09-01 to 09-05 outage postmortem, #550. One PR
because the task asked for one; the commits are split per fix, so review them
one at a time.

## Read this before merging

**Nothing here deploys on merge.** The heartbeat dispatcher and the poller
LaunchDaemon read scripts from disk at `/Users/jax/.behalfbot`, not from git.
That tree is currently dirty with the same five files #551 commits, so
repo-drift's auto-heal will keep refusing to fast-forward it until those local
edits are reconciled. Merging this and assuming the box is covered is exactly
the gap that produced root cause 3a.

**Expected first tick after it does land.** The existing
`scheduled-tasks/repo-drift-state.json` has no `last_alerted` field. If the
tree is still dirty and past quarantine, repo-drift will alert once more, then
go quiet for 24h. That is the new gate working, not a regression.

---

## Fix 1 - repo-drift re-alerted forever once quarantine opened

`scripts/gather-repo-drift.sh`

The #449 quarantine is an onset delay and nothing else. Once
`condition_age_hours >= QUARANTINE_HOURS` the `continue` stopped firing and the
repo was appended to `needs_sean` on every tick, forever. There was no
`last_alerted` field anywhere in the script.

Evidence, 2026-09-01 to 09-05: 514 runs, $45.79, 66% of all scheduled spend for
the window, 50+ near-identical posts into `#jax-devops`. The condition was a
dirty working tree only Sean could clear, and he was off-grid with no way to
stop it.

Adds `DRIFT_REALERT_HOURS` (default 24, env-overridable like the other knobs)
and tracks `last_alerted` per repo in the same state file. Same repo, same
reason, inside the window: no alert, but the ledger entry is still written so
`first_seen` and `last_alerted` carry forward.

Semantics preserved exactly:

- a clean repo drops out of the ledger and resets its clock
- a different reason earns a fresh quarantine window and a fresh re-alert
  window, because the alert that went out did not describe this problem
- errors are never gated. They do not enter this path at all.

`last_alerted` records emission, not delivery. This gather cannot see whether
the model actually posted, and per root cause 3c it often did not. Accepted
trade: the delivery-side dedup is the companion chassis PR's job. This one is
repo-drift-specific and belongs in the customer script.

## Fix 2 - the poller hammered Apple when its iCloud session died

`scripts/angel-location-poller.py`

`com.jax.angel-location-poller` runs on `StartInterval 600` with no backoff.
When the trust token expired on 2026-09-04 at 11:14, every tick did a full
authentication, exited fail-closed (`FAIL-CLOSED: 2FA required and no
interactive input available`), and pushed a new 2FA code that invalidated the
previous one. That ran unattended for about 25 hours. It made interactive
`--setup` nearly impossible, which is what Sean hit on 09-05, and repeated
failed authentications from one IP is what Apple rate-limits.

After 3 consecutive session-invalid results the poller writes `needs_setup`
into `~/.angel-vault/icloud-auth-state.json`, stops authenticating entirely,
and posts one alert to `#jax-ops` through the same webhook chain and message
shape `angel-burglar-alarm.py` already uses for poller health. The alert
carries the exact setup command, since the wrong one was in the docs. It stays
parked until `--setup` succeeds, which clears the flag; a successful poll also
resets the streak.

Why 3: one failure can be an Apple-side hiccup and giving up on the first would
park a working session. Three at 600s is 30 minutes and 3 pushed codes, which
is a dead session. It also matches `ANGEL_POLLER_FAIL_STREAK_N` in the alarm,
so both watchers agree on what "the feed is really down" means.

Only 2FA-required, 2SA-required and an outright rejected login advance the
streak. Network errors, timeouts and everything else do not - parking on a wifi
blip would silence the location feed for no reason. `authenticate()` now raises
`IcloudSessionInvalid` instead of exiting inline so the failure can be recorded
first; exit codes are unchanged (3 dead session, 2 missing or wrong-mode
credentials) with a new 4 for a parked run.

The fail-closed guarantee is intact and tested: `ICLOUD_PW_PATH` must exist and
be mode 0600 or the run exits 2 before parking is even consulted. Parking
suppresses re-authentication; it never relaxes anything.

**Alert interaction worth knowing:** a parked poller appends no pings, so the
alarm's existing "poller stale / job may be dead" alert also fires once. Two
one-time alerts, not a loop.

## Fix 3 - travel mode never suppressed the welfare check

`scripts/_dormant_gather.py`, `scripts/gather-dormant-sean.sh`, both prompts

Sean announced the roadtrip on 08-31 and `~/.angel-vault/mode.json` recorded it
correctly: `explicit_mode: "traveling"`, `set_reason` "not waking at home",
`auto_clear_at` 09-06. `explicit_mode` was read by exactly two files, both
belonging to the burglar alarm. Neither `gather-dormant-sean.sh` nor
`_dormant_gather.py` read `mode.json` at all. So the check ran as if he were
home and quiet:

| Day | Verdict | What happened | Under this PR |
|---|---|---|---|
| 09-01 | YELLOW, tier1=2 | pinged Sean in #jax | emits 0, model never invoked |
| 09-02 | YELLOW, tier1=2 | pinged Sean in #jax | emits 0, model never invoked |
| 09-03 | RED, tier1=3 | posted RED report, noting he was probably camping | 11:00 pre-nudge DMs him; noon posts RED only if he did not reply |
| 09-04 | YELLOW, tier1=2 | pinged Sean in #jax | emits 0, model never invoked |

The 09-03 row is the damning one: the model worked out he was camping and
posted anyway, because the verdict logic had no suppression path.

`_dormant_gather.py` now reads `mode.json`, honours `auto_clear_at`, and
surfaces `travel_mode` in the gather JSON. Reading is read-only on purpose:
`angel-mode-detector.py` deletes expired entries in place, but the vault is
bind-mounted read-only into the container this gather runs in, and a monitor
should not mutate the state it is judging.

Suppression zeroes the trigger fields. The dispatcher conditions are `threshold
tier1_quiet_count >= 2` (noon) and `>= 3` (11am pre-nudge), so zeroing
`tier1_quiet_count` and its deprecated `count` alias is what stops the model
being invoked at all - the cheap path and the quiet one. `verdict` and
`tier1_quiet_count_raw` still report what the signals actually said, and every
suppression writes a line to `logs/angel/dormant-sean.log`, so a silenced check
stays discoverable.

### Decision: RED still fires while travelling

It does, and the camera veto goes inert. Reasoning, also written into the code
next to the decision:

- **What happens if he has a stroke in a tent.** Phone dark, no Oura sync, no
  Discord gives tier1=3. At 11:00 the pre-nudge DMs him "are you OK". At noon,
  with no reply, the verdict is RED: in the current `monitor` mode that is a
  full RED report to Sean alone with the cascade disarmed; in `armed` mode the
  cascade fires. Time to detection is the same as it is at home.
- **What travelling changes is only the 2-of-3 day.** Being away from the flat
  and off the ring for a day is what a roadtrip is. Those days say nothing, and
  they produced four pings in five days.
- **Camera veto inert.** Motion at his empty flat is not evidence he is alive
  300km away. That is a burglar-alarm question and the burglar alarm asks it.
  Letting it rescue a travelling all-quiet day would mean a cat or a delivery
  could silence the one verdict that matters. Note the direction of this: while
  travelling, tier1=3 plus camera motion is RED where at home it would be a
  camera-rescue YELLOW. Travel mode makes the all-dark day MORE likely to
  escalate, not less.
- **Cost asymmetry.** A phone with no signal for a day looks identical to a
  phone whose owner is unconscious, so a travelling RED will sometimes be
  wrong. That costs one Discord message. A suppressed real RED costs the
  premise of the check, which exists because he lives alone and nobody else
  would notice. The 11am pre-nudge is the mitigation: he gets asked before
  anything escalates and one reply clears it.

Both prompts now carry the framing so a travelling RED is not misread:
`moved_today` quiet does NOT mean he stayed home (away from home that signal
reads active, so quiet means the location feed itself is dark), and the camera
must not be cited as reassurance.

`scheduled-tasks/dormant-sean-mode.json` is untouched. It remains the authority
over `dry_run` and its precedence rules are undisturbed.

## Fix 4 - stale `~/jax/` paths

Sean was handed a documented command on 09-05 that could not work:
`/Users/jax/jax/scripts/angel-location-poller.py --setup` gives `[Errno 2] No
such file or directory`.

Swept `~/jax/` and `/Users/jax/jax/` where the referent is the install root.
`HEARTBEATS.md` named `~/jax` as a watched repo; the gather watches
`$JAX_HOME`, which resolves to `~/.behalfbot`. The poller docstring is fixed in
the Fix 2 commit.

`docs/DR.md` needed both halves of its clone step fixed, not just the path - it
cloned `scrollinondubs/jax` into `~/jax`. A runbook that looks correct and
clones the wrong repo is worse than an obviously stale one. The rest of that
file was not audited.

`chassis.config.yaml`'s `state_path` was updated after checking it has no
consumer: nothing in the live chassis checkout at `/Users/jax/behalfbot` reads
the field (bootstrap.sh's config validation is still a TODO stub, and its list
of required fields does not include it).

### What was deliberately left alone

| Reference | Why |
|---|---|
| `/Users/jax/jax/dating-swipes` (both dashboard plists, dating prompts, skills) | real path, verified it exists and holds Sean's calibration screenshots |
| `~/jax/temp/` in `bfl-backfill.py` | real path |
| `~/jax-private/`, `~/jax-archive/` | unrelated, correct |
| `.env.example` generation note, `chassis.config.yaml` fork note | historical, dated |
| `docs/cutover/*`, `docs/research/*`, `scripts/cutover/2026-cutover-runbook.sh` | historical, dated |
| "Auto-migrated from `~/jax/.env` on 2026-05-02" in both migrate-env scripts | provenance stamp with the date baked in next to it |
| `repo-drift-alert-prompt.md` "cd ~/jax && git stash" | quoted as output the prompt must NOT produce |
| `services/shared/venice.py` rename rationale | explains why REPO is derived rather than hardcoded |
| SiYuan block `20260521034144-bki03vj` | Sean is handling it by hand |

---

## Test plan

- [x] `bash tests/test_repo_drift_realert.sh` - PASS, 13 checks. New harness:
      drives a throwaway git fixture through quarantined, alerts once, silent
      for three ticks inside the window, alerts again past it, a changed reason
      alerting immediately, and an uninspectable repo firing on every tick.
      `DRIFT_REPOS` exists so it never touches the live checkouts.
- [x] `python3.12 -m pytest tests/test_dormant_travel_mode.py
      tests/test_dormant_gather_verdict.py
      tests/test_angel_poller_auth_backoff.py
      tests/test_angel_location_freshness.py
      tests/test_dormant_prenudge_state.py -q` - 89 passed.
      Includes the pre-existing verdict and freshness suites, unchanged, so the
      `traveling` keyword did not disturb any existing caller.
- [x] New `tests/test_angel_poller_auth_backoff.py` (19 cases) covers: parks at
      exactly the threshold and not before, alerts once, further parked ticks
      never reach Apple and append no pings, a network error never advances the
      streak, and missing or 0644 credentials still exit 2 with the mode
      unrepaired. No iCloud, no network, no writes to the real vault.
- [x] New `tests/test_dormant_travel_mode.py` (25 cases) covers: the verbatim
      roadtrip `mode.json`, expired `auto_clear_at`, absent file, malformed
      JSON, a garbage timestamp, a non-traveling explicit mode, read never
      writes, camera rescue inert while travelling, and end to end through
      `main()` that a travelling YELLOW emits `count: 0` while a travelling RED
      emits 3.
- [x] `zsh -n scheduled-tasks/heartbeat-dispatcher.sh` - clean. That file is
      `#!/bin/zsh` and `bash -n` fails on its glob qualifiers on origin/main
      too, so zsh is the right linter for it.
- [x] `bash -n` on every other changed shell file, `ast.parse` on every changed
      Python file, `plutil -lint` on the changed plist.
- [x] Secret scan of the full diff: no tokens, keys or webhook URLs. The only
      credential-shaped string is a test fixture password.
- [x] No em dashes on any added line.

Not exercised: the live Apple auth path (deliberately - constructing
`PyiCloudService` is what pushes the 2FA code), the live Discord and iVideon
signals, and a real dispatcher tick. Those need the box.

Refs #550


jacketyjax:
## Two red checks, neither of them this PR

Both are worth explaining before anyone reads the check list as a verdict.

**`plugin-root-resolution` - pre-existing on `main`.** `test-plugin-root-resolution.sh` dies at line 262 with `CHASSIS_HOME: unbound variable`. The last two `main` runs of `shell-tests.yml` failed the same way (`8cc000b`, `1fcf6ae`), and it reproduces locally on the file untouched by this branch. `fix/plugin-root-test-unbound-chassis-home` is already in flight for it.

**`dco` - an artifact of a manual re-run, not a missing sign-off.** GitHub stopped delivering `pull_request` and `push` webhook events for this repo at about 12:47Z: the second commit (`afcc549`) landed on the remote and produced zero check runs. I re-ran the workflows by `workflow_dispatch` to get real results, and `check-dco.sh` reads `BASE` / `HEAD` from `github.event.pull_request`, which is empty on a dispatch, so it exits 2 with `check-dco: BASE and HEAD must be set` before checking anything. `dco-self-test` passed in the same run. Both commits do carry the trailer:

```
afcc549  Signed-off-by: Jax <jax@vibecodelisboa.com>
f3eb911  Signed-off-by: Jax <jax@vibecodelisboa.com>
```

Closing and reopening the PR did not re-fire the events either. It should settle on its own; if it has not by the time you look, a `--force-with-lease` no-op push or a fresh re-run once webhooks recover will give a genuine `dco` result.

**Everything else is green on `afcc549`**, including both new suites: `credential-expiry` (58 cases) and `claude-oauth-bridge-alert` (36 cases), plus `shellcheck`, `check-no-customer-state`, `diff-against-canonical`, `build` and the eleven other shell suites.

jacketyjax:
## Correction: I misdiagnosed the missing checks above, and all 20 now pass

My earlier comment blamed dropped GitHub webhooks. That was wrong, and the way it was wrong is worth writing down.

**The real cause: this PR was conflicting with `main`.** `#195` (the kill switch) and `#197` (the plugin-root test fix) merged while this branch was open, and both touched `chassis/CHANGELOG.md` and `shell-tests.yml`. GitHub runs `pull_request` workflows against the merge ref, and it cannot compute one for a conflicting PR, so it silently ran nothing. `gh pr checks` reported "no checks reported", which reads identically to a webhook outage.

What should have told me apart: `pull_request` events were firing normally for other branches the whole time (`fix/plugin-root-test-unbound-chassis-home` at 12:59Z, `feat/dispatcher-fire-caps` at 13:17Z). I checked repo-wide Actions health and read "other runs are happening" as reassurance instead of asking why those and not this one. `gh pr view --json mergeStateStatus` would have said `DIRTY` at any point. One field, thirty seconds.

Also wrong in that comment: a `--force-with-lease` no-op push generates no event at all, so it would not have helped either.

**Now resolved.** Merged `origin/main` in (`a820dd3`) rather than rebasing, keeping both sides of every conflict:

- `chassis/CHANGELOG.md` - my two `### Added` entries and my `### Fixed` entry sit alongside the kill switch's, in the right sections. The kill-switch entry stays first.
- `.github/workflows/shell-tests.yml` - `control-flags`, `credential-expiry` and `claude-oauth-bridge-alert` all present, with both path lists merged.
- `chassis/VERSION` - **my bump is now a no-op.** `#195` already moved it to `0.6.0`, and since that release is still unreleased this change rides the same `0.6.0` under `## Unreleased`. The update checker delivers it as changelog drift either way (#147), so nothing is stranded.

`b1e02ce` is an empty commit pushed while I still believed the webhook theory. Its message is wrong for the same reason; it is harmless, and dropping it would need a force-push.

**All 20 checks green on `a820dd3`**, including `dco` (a genuine `pull_request` run this time - the earlier red was my `workflow_dispatch`, where `check-dco.sh` gets empty `BASE`/`HEAD`) and `plugin-root-resolution` (fixed by `#197`). Both new suites pass: `credential-expiry` 58 cases, `claude-oauth-bridge-alert` 36 cases. Locally, `test-control-flags.sh` (30) and `test-plugin-root-resolution.sh` (44) also pass against this branch, so the merge did not disturb what just landed.

PR #195, opened 2026-09-05
feat(chassis): remote kill switch that works with no shell and no Claude

jacketyjax:
## What this is

A remote kill switch handled inside a standalone bridge process, so it works when Claude is dead, when the dispatcher is wedged, and when there is no shell.

**This is the fix that would have ended new-jaxity#550 on day one.**

## Why

From `scrollinondubs/new-jaxity#550`. Between 2026-09-01 and 2026-09-05 the reference install's operator was off-grid camping. A heartbeat called `repo-drift` fired every 30 minutes on a condition that could never self-clear, because only he could clear it:

- **514 model invocations**
- **$45.79**, 66% of all scheduled spend for the period
- **50+ near-identical messages** into one Discord channel
- **4 days**

In the same window his Tailscale node key hit its 180-day expiry (no ssh, no VNC) and the macOS Keychain lost the Claude Code OAuth token, so the host `claude` session that serves Discord was dead and replies reached nobody. He could watch the alerts land on his phone and had **no way to stop any of it**. He drove home and pulled the power cord out of the Mac Mini.

Conservation mode already existed and would have stopped most of that spend. It is flipped by `scripts/conservation-mode.sh`, which needs a shell, which is the one thing he did not have.

**Every stop path the chassis had shared a dependency with something that was down.** That is the defect this PR fixes. The runaway heartbeat is a separate issue.

## What changed

**`chassis/scripts/discord-control-listener.py`** - a standalone process that polls the Discord REST API with the bot token, matches a fixed lexicon from the install's principal, writes the dispatcher's flag files directly, and acks in-channel. No model in the path, no shell in the path.

| Command | Aliases | Effect |
|---|---|---|
| `halt [<duration>] [<reason>]` | `mute`, `silence` | Skip EVERY heartbeat, `critical` included |
| `resume` | `unmute`, `unhalt` | Clear halt AND conservation |
| `conserve [<duration>] [<reason>]` | `throttle`, `conservation on` | Skip `normal` + `background` only |
| `unconserve` | `conservation off` | Clear conservation |
| `status` | | Report both flags |

Whole-message match, case-insensitive, leading bot mention and trailing punctuation stripped. "we should probably halt the drift alert" does nothing. `stop` is deliberately not in the lexicon - people type it mid-conversation to interrupt.

**New `scheduled-tasks/halt.json`**, checked before conservation mode and ignoring `criticality` entirely. Conservation lets `critical` through, which is right for a quota throttle and wrong for an operator asking for silence. Same five-key schema as `conservation-mode.json` rather than a parallel one, so `backup-to-s3.sh`, `gather-quota-check.sh` and `scripts/conservation-mode.sh` need no new vocabulary.

**Omitting the duration means no auto-lift.** An auto-resume that fires while the operator is still off-grid restarts the outage, and by definition they are not there to see it.

**Authorisation** reuses `CHASSIS_PRINCIPAL_USER_ID` (the principal-policy hook's id), falling back to `INSTALLER_DISCORD_USER_ID`. That hook fails **open** on a missing id, correctly. This one fails **closed**: refuses to start, exits 78, and the entrypoint writes `remote kill switch UNARMED` plus the reason into the container log. `bootstrap-audit.sh` gains Gap 6 for the same condition.

**It runs as a supervised child of the dispatcher loop, not a new compose service.** Existing installs gain the switch by pulling an image, with no customer-side compose edit - which matters more here than the isolation a separate container would buy. A `control-listener` entrypoint mode exists for installs that want it isolated anyway. The child is deliberately not waited on: a wedged tick must not take the switch down with it.

### Also fixed: conservation mode's auto-lift has never worked inside the container

`is_conservation_mode()` parsed `auto_lift_after` with `date -j -f ... || echo 0`. `date -j` is BSD only, so on Debian (every containerized install) the parse always failed, `lift_epoch` became 0, `now >= 0` was always true, and **any `conservation-mode.json` carrying a non-null `auto_lift_after` was disabled on the very next tick**. It logged `CONSERVATION - auto-lift triggered`, which is exactly what a correct auto-lift logs, so nothing could surface it. Found while building the halt flag, which needed the same arithmetic and would have inherited the same bug.

The parse moves to `chassis/scripts/_control-flags.sh`, shared by both flags. **An unparseable timestamp now keeps the flag ON rather than lifting it** - for a kill switch the direction of the failure matters more than the failure.

## Known caveat: the reference install is UNARMED today

`CHASSIS_PRINCIPAL_USER_ID` and `INSTALLER_DISCORD_USER_ID` are both unset in the reference install's container. Gap 6 fails against it right now, and the listener will log `UNARMED` and exit 78 on the next image pull.

The customer-side follow-up is one line in `.env`, a re-bake and a container restart:

```
CHASSIS_PRINCIPAL_USER_ID=<principal Discord snowflake>
```

Until that lands, this PR ships the gate and the audit finding but not a working switch on that install. Calling that out explicitly rather than shipping something that silently never arms, which is the failure shape #550 is about.

## What halt does NOT stop

Documented in `docs/remote-kill-switch.md` and repeated verbatim in the in-channel ack, because in the moment the ack is the only documentation the operator has:

- An in-flight `claude -p`. The gate is at tick start; a running tick finishes, up to ~40 min with retries.
- Host-side launchd / systemd jobs: the iCloud poller, `discord-restart`, `discord-watchdog`, plugin agents.
- A live interactive Claude session.
- The listener itself, which has to stay up to hear `resume`.

## Test plan

- [x] `bash chassis/scripts/test-control-flags.sh` - 30 passed, 0 failed. Stubs `date` on PATH to force each dialect, runs the full matrix three times, and asserts the PRE-fix idiom still misparses a future expiry under GNU date so the discrimination cannot rot.
- [x] `pytest chassis/scripts/tests/` - 129 passed (49 new). Stdlib only, no new CI dependency.
- [x] `shellcheck --severity=error` clean on `_control-flags.sh`, `test-control-flags.sh`, `bootstrap-audit.sh`, `backup-to-s3.sh`, `entrypoint.sh`.
- [x] `zsh -n` on the dispatcher, `bash -n` on the shell scripts, `ast.parse` on the listener.
- [x] **Live Discord round trip.** Listener run from this worktree against a scratch `CUSTOMER_HOME` (never the live one), real bot token, real channel. A sentence containing "halt" was correctly ignored; `halt kill-switch e2e test` wrote `halt.json` with `enabled: true`, `reason: "kill-switch e2e test"`, `auto_lift_after: null` and posted the ack; `status` reported `halt: ON since ... - reason: ... - auto-lift: none`; `resume` cleared both flags and posted `RESUMED`. Cursor persisted, no command reapplied on the next poll.
- [x] **Dispatcher gate, real dispatcher, scratch dir.** With `halt.json` enabled, a `criticality: critical` heartbeat did not run and its gather never executed; the log shows `HALTED - every heartbeat skipped`. With `auto_lift_after` in the past, the same heartbeat ran, the file self-cleared to the five-key "off" document, and the log shows `HALT - auto-lift triggered, flag cleared`.
- [x] Gap 6 run against the reference install's `CUSTOMER_HOME`: fails with `no principal user id - this install cannot be halted from Discord`, which is the correct verdict today.
- [ ] Not verified: the entrypoint supervisor path in a built image. It is shell that shellchecks clean and the file-copy is covered by the existing `COPY chassis/ /app/chassis/`, but no image was built for this PR.

Refs `scrollinondubs/new-jaxity#550`.

Do not merge without review.


jacketyjax:
CI note: `plugin-root-resolution` fails on this PR and also fails on `main` with the identical error - `test-plugin-root-resolution.sh: line 262: CHASSIS_HOME: unbound variable` (run 33300296526, 2026-08-30, branch main). Pre-existing and unrelated to this change; nothing here touches `resolve-plugin-root.sh` or its suite.

The first push of `control-flags` did fail and is now fixed: the date stub forwarded `date -j` to the real binary, so forcing the BSD dialect was impossible on a GNU-only runner. Both dialects are emulated now. 30/30 on macOS, 30/30 on debian:bookworm-slim, plus the real GNU date path checked without the stub.

PR #196, opened 2026-09-05
feat(dispatcher): daily fire caps, alert dedup and a suppression ledger

jacketyjax:
> **Do not merge.** Sean ratifies.

Refs scrollinondubs/new-jaxity#550.

## What happened

Between 2026-09-01 and 09-05, with Sean off-grid, `repo-drift` fired **514 times in five days for $45.79 - 66% of all scheduled spend** ($69.45 total) and posted 50+ near-identical alerts into a channel nobody could reach.

```
repo-drift        runs= 514   $45.79      <-- 66% of ALL scheduled spend
telegram-groups   runs=  39   $ 7.58
pulse-triage      runs=   4   $ 4.16
TOTAL                         $69.45
```

Nothing was broken. The gather correctly reported a dirty working tree, the condition correctly evaluated true, and the auto-heal correctly refused to touch a dirty tree. Five files sat uncommitted for a week, so `count > 0` came back on every tick, forever, and only a human could clear it.

`budget: 0.10` was configured and was never once exceeded, because `budget` is a per-invocation ceiling and per-invocation ceilings do not bound a loop. 514 runs at about $0.089 each.

## What this adds

Four optional heartbeat fields, all opt-in.

| Field | Example | Bounds |
|---|---|---|
| `max_fires_per_day` | `3` | Model invocations per calendar day |
| `max_usd_per_day` | `0.50` | Spend per calendar day, read from the dispatcher's own telemetry |
| `realert_after` | `24h` | How long an unchanged condition stays quiet after alerting |
| `dedupe_key` | `.needs_sean \| map({repo, reason})` | jq expression naming the condition's identity |

`repo-drift` with the fields it should have had turns 514 runs and 50+ messages into one run and one message per day, until someone commits the five files.

### Caps are absolute

Once one trips, that heartbeat invokes no model again until the date rolls. A change in the condition does not lift it: the condition here changes constantly (the ages climb) while the problem does not, and a cap with an escape hatch is not a cap. A failed fire counts against the ceiling, because it spent tokens; a failed fire does not start the dedup window, because it reached nobody.

The dollar cap reads telemetry rather than a counter of its own. Telemetry is the surface you audit after the fact, and a cap that disagreed with it would be worse than no cap. It is checked pre-fire, so a day can overshoot by at most one run's `budget`.

### How the fingerprint stays stable

This is the part that fails silently if you get it wrong, and this repo has been bitten by it.

`repo-drift`'s alerts read "unresolved for 4 days", then "5 days", then "131h", then "180 hours". Its gather emits `condition_age_hours`, `dirty_oldest_age_hours` and `ahead_newest_age_hours` next to the actual finding. A content hash over the gather output, or over the rendered alert, never matches itself: the cooldown never fires and the only symptom is that nothing changed.

So the identity is never derived from the payload by guesswork:

1. **`dedupe_key`**, a jq projection, when set. Project what is wrong, leave out every age, timestamp and counter.
2. **Otherwise, for a `threshold` condition**, the value the condition tested - the number the dispatcher made its own fire decision on. Extracted by a `threshold_actual()` now shared with `evaluate_condition`, so the decision and the identity cannot drift apart on how a payload is read. Coarse by design: same count, different underlying problem, same identity. That is exactly when you reach for `dedupe_key`, and there is a test asserting the coarseness so it is documented rather than discovered.
3. **Otherwise, nothing.** `always` and `ask_model` with no `dedupe_key` log a warning and fire.

**A strip-the-volatile-looking-field-names default was considered and rejected.** Guessing which keys move from their names is the same bug with better manners, and on this exact payload it would have hashed `behind` and `ahead`, which move whenever upstream moves.

Runtime detector for a bad key: if the fingerprint changes on three consecutive fires (`DEDUPE_CHURN_WARN_THRESHOLD`), the dispatcher logs a warning naming the heartbeat. A dedup that suppresses nothing must not be able to look like a dedup with nothing to suppress.

The dedup identity is cleared when the condition goes false, so the same problem recurring next month is news again.

### Quiet, but discoverable: the suppression ledger

The same outage had a credential-sync bridge that detected the real problem and logged it 288 times over four days into a file nobody reads. Muting an alert into total silence rebuilds that.

Three surfaces:

1. Every suppressed tick logs `CAPPED <name>` / `DEDUPED <name>` and sets `last_result` to `fire_capped` / `alert_deduped` in `heartbeat-state.json`.
2. A tripped cap posts **one** notice per heartbeat per day: the arithmetic, the last verdict, how many checks have been swallowed, and how to lift it. The remedy is branched on which cap tripped, because clearing `fires_today` does nothing to a ceiling measured from telemetry.
3. **The ledger**, for the weekly rollup.

#### Ledger contract (stable, for the customer-side weekly rollup)

**Path:** `${CUSTOMER_HOME}/scheduled-tasks/heartbeat-suppressions.json`, beside `heartbeat-state.json`. One record per heartbeat, keyed by name, updated in place.

```json
{
  "repo-drift": {
    "reason": "fire_cap",
    "detail": "max_fires_per_day reached: 3 of 3 model invocations already spent today",
    "fingerprint": "9f2c8e...",
    "first_suppressed_at": "2026-09-01T00:14:11",
    "first_suppressed_epoch": "1788221651",
    "last_suppressed_at": "2026-09-05T13:44:02",
    "last_suppressed_epoch": "1788615842",
    "suppressed_ticks": 471,
    "last_decision": "YES - count=1 (> 0)"
  }
}
```

| Field | Meaning |
|---|---|
| `reason` | `fire_cap`, `usd_cap` or `dedup` |
| `detail` | One line of human-readable arithmetic or window |
| `fingerprint` | Condition identity, empty when dedup is not in play |
| `first_suppressed_at` / `_epoch` | Start of this unbroken quiet period |
| `last_suppressed_at` / `_epoch` | Most recent suppressed tick |
| `suppressed_ticks` | Checks swallowed since `first_suppressed_at` |
| `last_decision` | Condition verdict at the last suppressed tick |

Both an ISO string and an epoch are stored, because `date -j -f` is macOS-only and a consumer that has to parse the ISO form to do arithmetic is not portable.

**Entries clear when the condition clears** - on a false gather, or on a fire that actually goes out. That is what makes "quiet because it was fixed" distinguishable from "quiet because it was muted".

**Reader:** `chassis/scripts/gather-heartbeat-suppressions.sh`. A file read and one `jq` pass, no model call.

```
$ scripts/gather-heartbeat-suppressions.sh --min-hours 24
{"count":1,"generated_at":"...","ledger_path":"...","suppressed":[
  {"heartbeat":"repo-drift","reason":"fire_cap","detail":"...","fingerprint":"9f2c8e...",
   "first_suppressed_at":"2026-09-01T00:14:11","last_suppressed_at":"2026-09-05T13:44:02",
   "suppressed_ticks":471,"quiet_hours":133}]}
```

It emits `count` and always exits 0, so it works unchanged as a heartbeat gather with `condition: threshold count > 0`. A ready-to-copy weekly-rollup block is in `HEARTBEATS.md.template`, in the container path shape.

`quiet_hours` is derived at read time and is deliberately not stored and not part of any fingerprint: a duration that grows every tick is precisely the value that must never reach a content hash.

## Verification

Everything below was run on this branch, on macOS zsh 5.9.

```
$ zsh chassis/scheduled-tasks/tests/test_fire_caps.sh
PASS: 68 assertions

$ zsh chassis/scheduled-tasks/tests/test_fire_caps_e2e.sh
  spend: $1.0692 across 12 runs (uncapped would have been $2.1384 across 24)
PASS: 20 assertions

$ zsh chassis/scheduled-tasks/tests/test_state_write_race.sh
PASS: 40 concurrent distinct-key writers all survived, no leftover tmp files

$ shellcheck -S error chassis/scripts/gather-heartbeat-suppressions.sh chassis/scripts/backup-to-s3.sh
(clean)

$ zsh -n chassis/scheduled-tasks/heartbeat-dispatcher.sh
(clean)
```

**The suites are proven to discriminate, not merely to pass.** The e2e suite stubs `claude` on PATH and drives six real dispatcher ticks over a gather that always reports work, with four heartbeats: one capped by fires, one by dollars, one deduped, one carrying none of the new fields. Run against `origin/main`'s dispatcher it produces **24 model invocations and $2.1384**; on this branch **12 and $1.0692**, with 15 assertions failing against `origin/main`:

```
$ git show origin/main:chassis/scheduled-tasks/heartbeat-dispatcher.sh > /tmp/old.sh
$ # ... e2e suite pointed at /tmp/old.sh
FAIL: max_fires_per_day: 2 must produce exactly 2 model invocations across 6 ticks (expected '2', got '6')
FAIL: max_usd_per_day: 0.20 must stop the heartbeat on the tick after telemetry crosses it (expected '3', got '6')
FAIL: an unchanged condition must alert once inside realert_after ... (expected '1', got '6')
FAIL: 4 heartbeats x 6 ticks is 24 runs uncapped; caps and dedup must bring it to 12 (expected '12', got '24')
  spend: $2.1384 across 24 runs
```

The fingerprint-stability case in the unit suite feeds two real `repo-drift` payloads differing only in `condition_age_hours` (96 vs 180) and `dirty_oldest_age_hours`, asserts one fingerprint, and carries a **control assertion that a naive hash of the same pair does differ** - so the stability assertion cannot pass by hashing a constant. Backwards compatibility is asserted by running it: the e2e suite's unconfigured heartbeat must fire on all six ticks.

CI: new `heartbeat-fire-caps` job in `shell-tests.yml` installs zsh (the runner image does not guarantee it) and runs all three suites, including the previously CI-less `test_state_write_race.sh`. Both `paths:` lists updated. The two new zsh suites are added to `shell-lint.yml`'s `ignore_names`, as shellcheck does not support zsh.

## Deliberate decisions and deviations

- **`realert_after` is the opt-in switch, not a default 24h.** The brief said "default 24h"; backwards compatibility is a hard rule, and a dedup that turns itself on would change the behaviour of every existing heartbeat on every install. 24h is documented as the recommended value instead.
- **Caps are absolute rather than "until the day rolls or the condition changes".** Dedup owns condition identity; a cap that any condition change lifts would not have stopped `repo-drift`, whose payload changes every tick.
- **`chassis/VERSION` bumped 0.5.1 -> 0.6.0, which deviates from this repo's release convention.** VERSION here moves only in a dedicated `chore(release):` commit that also retitles `## Unreleased` (see c312173). The brief asked for the bump, so it is here, with my entry left under `## Unreleased` rather than cutting a release that would sweep in other people's merged work. **If you would rather keep the convention, drop the `chassis/VERSION` hunk and cut 0.6.0 separately.**
- **No separate reporting CLI beyond the gather script.** The gather is the queryable surface; a second renderer would be one more thing to keep in sync.
- **The `decision='...'` / `attempt=1` noise in the dispatcher log** is a pre-existing zsh bare-`local` re-declaration quirk that the file's own comments document. Left alone: out of scope, and it is only log noise. Every `local` added here is initialized.
- **Not done: a customer-side weekly-rollup heartbeat.** That is Sean's to wire once this lands, which is why the ledger contract is written down rather than assumed.

## Customer-side follow-up

`.gitignore` here covers the chassis repo. **The customer install needs the same entries added to its own `.gitignore`**, or the first suppression writes an untracked file under `scheduled-tasks/`, which is a `repo-drift` watched path, and the runaway heartbeat this PR fixes will page about the file the fix wrote:

```
scheduled-tasks/heartbeat-suppressions.json
scheduled-tasks/heartbeat-suppressions.json.*.tmp
scheduled-tasks/heartbeat-suppressions.json.*.clr
scheduled-tasks/heartbeat-suppressions.json.lock/
```

## Files

- `chassis/scheduled-tasks/heartbeat-dispatcher.sh` - caps, dedup, fingerprint, ledger; `_state_lock_acquire/release` now take an optional target so the ledger reuses the state file's proven mutex instead of a second copy of it
- `chassis/scripts/gather-heartbeat-suppressions.sh` - new, the ledger reader
- `chassis/scheduled-tasks/tests/test_fire_caps.sh` - new, 68 assertions
- `chassis/scheduled-tasks/tests/test_fire_caps_e2e.sh` - new, 20 assertions
- `chassis/HEARTBEATS.md.template` - schema rows plus a "Runaway control" section with the ledger contract
- `docs/heartbeat-dispatcher.md` - the why, and the updated flow diagram
- `chassis/CHANGELOG.md`, `chassis/VERSION`, `chassis/scripts/backup-to-s3.sh`, `.gitignore`, `.github/workflows/{shell-tests,shell-lint,no-customer-state-in-chassis}.yml`


jacketyjax:
`plugin-root-resolution` is red on this PR and is **not from this branch**. It fails identically on `origin/main` (`7720ee3`), checked out clean and run locally:

```
FAIL [s8 falls back to baked] expected '.../s8/baked', got ''
chassis/scripts/test-plugin-root-resolution.sh: line 262: CHASSIS_HOME: unbound variable
```

Scenario 12 (added in 1fcf6ae, #178) reads `$CHASSIS_HOME` in its `check` description string while the suite runs under `set -u` and `fresh_env` has unset it, so the suite dies before printing its summary. Scenario 8 fails just above it. Nothing in this PR touches `resolve-plugin-root.sh`, its test, or `CHASSIS_HOME` outside the dispatcher. Worth its own issue.

`heartbeat-fire-caps` (the new job, all three zsh suites) passes on the Linux runner.

PR #199, opened 2026-09-05
ci(shell): run the shell suites on macOS bash 3.2 with a BSD userland

jacketyjax:
## Why

Chassis code runs in two places. The Debian container is bash 5.x with GNU coreutils. The operator's macOS host is `/bin/bash` **3.2.57** (Apple has shipped 3.2 since 2007 for licensing reasons) with a BSD userland. The repo's own comments already say "host is macOS/BSD, container is Linux". CI only ever checked one of them.

Three bugs shipped **green** through the ubuntu-only lane on 2026-09-05 (new-jaxity#550) and broke on contact with a real machine:

1. `${BAKED_ROOTS[-1]:-}` in `chassis/scripts/resolve-plugin-root.sh`. Negative array subscripts need bash 4.3. On 3.2 that is a `bad array subscript` error that expands to empty, so the resolver handed callers an EMPTY plugin root on the compose-failure path - the exact silent no-op its own comment forbids. Fixed in #197.
2. `stat -f %m` in the colima work. A *filesystem* query on GNU coreutils, an mtime query on BSD. Chained with `||` the two outputs concatenated and silently disabled a recycled-pid lock breaker. Fixed in #193.
3. `date -j -f` in `is_conservation_mode()`. BSD only, so inside the Debian container `lift_epoch` became 0, `now >= 0` was always true, and any conservation flag with an `auto_lift_after` was lifted on the very next tick while logging the same line a working auto-lift logs. Fixed in #195.

Note the direction of 3: it broke on **Linux** because the code was BSD-only. So this is not "add macOS coverage", it is "both target platforms must be exercised". Bug 1 is the only one of the three that actually needs bash 3.2. Every check added here is symmetric.

## What changed

| File | Change |
|---|---|
| `.github/workflows/shell-tests.yml` | New `macos-bash32` job. Blocking. |
| `.github/workflows/shell-lint.yml` | New `portability` job: blocking self-test + advisory scan. |
| `.github/scripts/check-shell-portability.sh` | New. Advisory grep for single-platform shell constructs. |
| `.github/scripts/test-check-shell-portability.sh` | New. 30-assertion self-test for the above. |
| `chassis/scripts/test-control-flags.sh` | One trailing `# portable-ok:` on the deliberately pre-fix BSD idiom. |
| `chassis/CHANGELOG.md` | Entry under `## Unreleased`. `chassis/VERSION` untouched. |

### The interpreter pin, which is the load-bearing part

The suites invoke their subjects as `bash "$SCRIPT"`, and half the subjects carry `#!/usr/bin/env bash`. Both resolve bash through `PATH`. GitHub's macOS image ships Homebrew, so `PATH` may well hand out bash 5.x - and then this entire lane would be a second ubuntu lane wearing a macOS hat: green for the wrong reason, which is precisely the failure class the postmortem is about.

So the job creates a shim dir with `bash -> /bin/bash`, prepends it to `PATH`, and then a **separate** step (`GITHUB_PATH` only applies to later steps) asserts, failing the job on any of them:

- `bash`, `/usr/bin/env bash` and `/bin/bash` all report `3.2.x`
- a negative array subscript is genuinely broken here (compares the **value**, not the exit code - `echo "${a[-1]}"` still exits 0 on 3.2, which is exactly why bug 1 was silent)
- `stat` accepts BSD `-f` and rejects GNU `-c`
- `date` accepts BSD `-j` and rejects GNU `-d`

### It counts assertions, not exit codes

Each suite's output is parsed for its summary line (`N passed` or `PASS: N`). A suite that exits 0 having reported **zero** assertions fails the job. The run must also total at least 400 assertions (it totals 434 today) to guard against a broken checkout or a loop that iterates over nothing. A CI job that is green because it silently ran nothing is the thing being fixed here.

## Which suites run on macOS

Fifteen of the sixteen the ubuntu lane runs. Parity is deliberate: a divergence between the two lists is how one lane quietly stops covering something. Thirteen are bash; two are from the zsh `heartbeat-fire-caps` job that #196 landed on main while this branch was open, and unlike the ubuntu job this lane needs no install step for them because macOS ships zsh 5.9.

| Suite | Assertions on macOS |
|---|---|
| `test-chassis-update-health.sh` | 16 |
| `test-plugin-root-resolution.sh` | 44 |
| `test-chassis-root-health.sh` | 8 |
| `test-container-liveness.sh` | 7 |
| `test-chassis-root-resolution.sh` | 35 |
| `test-bootstrap-audit-memory.sh` | 12 |
| `test-memory-canary.sh` | 18 |
| `test-chassis-update-check.sh` | 46 |
| `test-chassis-update-mode.sh` | 54 |
| `test-discord-reply-param-fix.sh` | 24 |
| `test-colima-ensure.sh` | 31 |
| `test-control-flags.sh` | 30 |
| `plugins/dating/tests/test-emulator-state.sh` | 40 |
| `chassis/scheduled-tasks/tests/test_state_write_race.sh` (zsh) | 1 |
| `chassis/scheduled-tasks/tests/test_fire_caps.sh` (zsh) | 68 |
| **total** | **434** |

None needed a Docker daemon (they all stub the docker CLI) and none needed a GNU-only binary. `jq`, `python3` and `git` are the only external deps and all three are on the runner image.

**Excluded, same as the ubuntu lane:** `test-daily-log-gather.sh`, `test-compose-verify.sh`, `test-merge-plugin-triggers.sh`, `test-validate-heartbeats.sh`. These are not wired into CI on **either** platform today. Adding them is a separate change - it is a gap in the ubuntu lane first, and fixing it here would hide that.

**Excluded, macOS only: `chassis/scheduled-tasks/tests/test_fire_caps_e2e.sh`.** It is excluded because it *found something*, not because it was inconvenient. See below.

## The one existing failure, and what it is

The thirteen bash suites passed on bash 3.2.57 / macOS 26.3.1 on the first run, which is the good outcome - #193, #195 and #197 had already fixed the three known breakages.

Then #196 landed `heartbeat-fire-caps` on main while this branch was open, I added its three zsh suites for parity, and `test_fire_caps_e2e` ran for **over eight minutes** on a `macos-latest` runner against 22 seconds on ubuntu ([run 33972207796](https://github.com/scrollinondubs/behalfbot/actions/runs/33972207796), step started 14:34:24, still running at 14:42:45 when I cancelled it).

Reproduced on my own machine with nothing but `PATH` changed:

| PATH | Result |
|---|---|
| normal (Homebrew present) | passes in 42s |
| `PATH=/usr/bin:/bin:/usr/sbin:/sbin` | hangs, killed at 100s |

`ps` during the stock-PATH run shows the dispatcher parked on its retry sleeps. The cause is `chassis/scheduled-tasks/heartbeat-dispatcher.sh:77`:

```zsh
TIMEOUT_CMD="$(command -v gtimeout 2>/dev/null || command -v timeout 2>/dev/null || echo timeout)"
```

Both are GNU coreutils. Neither exists on a stock macOS install, so `TIMEOUT_CMD` becomes the literal string `timeout`, `$TIMEOUT_CMD 1200 /bin/zsh -c` around the `claude -p` call returns 127, and the retry policy sleeps 20s then 60s per attempt before opening the failure circuit. **On a fresh Mac install no heartbeat can invoke a model at all**, and the only clue is a `command not found` on stderr. Invisible until now because the reference install has Homebrew on PATH.

**Classification: (c), its own issue.** Filed as #203 with the reproduction. Fixing the dispatcher's watchdog resolution is a separate change with real blast radius, and it is not `brew install coreutils` - that turns a bug into a documented prerequisite and removes the exact userland this lane exists to exercise. The suite goes back on the macOS list the day #203 is fixed, and the workflow comment says so.

Two things this PR does add as a result, both of which should have been there from the start:

- **A 180-second per-suite watchdog**, hand-rolled, because GNU `timeout` is precisely the binary this platform does not have. A killed suite is reported as `HUNG` rather than as a generic failure, and its partial output is printed - the previous shape only `cat`'d the log after the suite exited, so a hang produced no output at all.
- **`timeout-minutes: 15`** on the job as a backstop.

Finding a hang now costs three minutes instead of a six-hour runner.

## Proof the lane discriminates

Reintroducing bug 1 (`fallback_root="${BAKED_ROOTS[-1]:-}"` in place of the arithmetic-index spelling):

```
bash 5.2.15 / Debian, non-root:   test-plugin-root-resolution: 44 passed, 0 failed   -> GREEN
bash 3.2.57 / macOS 26.3.1:       test-plugin-root-resolution: 43 passed, 1 failed   -> RED
  FAIL [s8 falls back to baked] expected '.../s8/baked', got ''
```

That is the bug reproducing as an empty plugin root, caught by exactly one of the two lanes. Reproduced on CI too: on commit `9399658`, `plugin-root-resolution` (ubuntu) passed and `macos-bash32` failed on `s8 falls back to baked` with an empty root, on the same commit. Full evidence in the comment below.

## Cost

Locally the fifteen suites take **29 seconds** wall clock on an M4. On the runner, image spin-up dominates. The job runs in parallel with all fourteen ubuntu jobs, so it adds no serial time to a PR unless it becomes the slowest job in the matrix; measured numbers from this PR's own runs are in a comment below. macOS runners bill at 10x, but this repo is public so Actions minutes are free and the multiplier costs nothing.

One loop rather than one job per suite: each extra macOS runner would pay the image spin-up again to run a suite that mostly takes under three seconds. Every suite runs even when an earlier one fails, so one red suite cannot hide the rest.

## The lint: worth it, with teeth

The counter-argument on record is that behavioural coverage beats grepping for `[-1]`. That is correct, and it is why `macos-bash32` is the main event and this is `continue-on-error`. But it did not prevent bugs 2 and 3, and it never will for the code the suites do not reach: hooks, gather scripts, one-shot migration helpers, the zsh dispatcher. Nobody is going to write a behavioural suite for a migration script that runs once per install, and the failure mode there is silence, not a crash. shellcheck runs at `severity: error` and knows nothing about which bash version or which userland this code must survive, so it had nothing to say about any of the three bugs.

Precision measures, so it does not become noise:

- Comment-only lines are skipped, and the check excludes itself and its own fixtures.
- `date` and `stat` must appear in a **command position** (start of line, or after `(` `|` `;` `&` `{`), so the words `date -j` inside a quoted failure message do not warn. `test_fire_caps.sh:491` does exactly that, and there is a regression fixture for it.
- Constructs are anchored to shell syntax, so python `items[-1]` inside an inline program does not match - only `${NAME[-N]}` does.
- **Dialect pairs are matched over a 7-line window**, so the ordinary two-line "try one dialect, fall back to the other" idiom stays silent. `_control-flags.sh`, `gather-local-health.sh`, `colima-ensure.sh` and `encrypted-s3-upload.sh` all pass clean.
- For `stat` the window checks **order**, not just presence: GNU `stat -f %m FILE` prints a filesystem block to **stdout** and *then* exits 1, so a BSD-first chain pollutes the output of the GNU fallback behind it. That is bug 2 exactly, and GNU-first is the only safe order.
- GNU-only binaries are suppressed by a `command -v` probe in the window, which is how the repo already guards most of them.
- Escape hatch: a trailing `# portable-ok: <which platform, and why>`.
- Only file extension `.sh` plus **first-line** shebang matches, so fenced bash blocks inside markdown docs are not scanned.

It runs under both `gawk` and `mawk` with identical output.

**The self-test is blocking even though the scan is not.** 30 assertions: the three bugs verbatim plus every other bash-4 construct must fire, and the *fixed* spellings already in the tree (`${A[$(( ${#A[@]} - 1 ))]}`, GNU-first stat, two-line date fallbacks, `declare -a`, `command -v` guards, `sed -i.bak`) must produce **zero** findings. It also asserts the check reports how many files it scanned, so it cannot pass by walking an empty list.

**Findings on the tree as it stands: 12, every one pre-existing and every one real.** They are filed as #200, #201 and #202 rather than fixed here, one PR per concern. Two are worth calling out because they are live and they are the same class as the three above:

- `chassis/scheduled-tasks/heartbeat-dispatcher.sh:389` parses `last_checked` with `date -j -f ... || echo 0` and has no GNU fallback, so inside the container `last_epoch` is 0 and **every `every Nh` heartbeat matches on every tick**. Verified against the reference install's own dispatcher log rather than inferred: `books-import` is `schedule: every 6h` and matched on all 45 ticks of 2026-09-05, at 00:03, 00:24, 00:42, 01:00, 01:17, 01:34, 01:50, 02:06.
- `chassis/scripts/migrate-from-old-clone.sh:146` uses `mapfile`, which does not exist on bash 3.2, in a host-side migration script. `TOP_LEVEL` comes back empty, the dynamic exclude list is empty, and the migration clobbers chassis-tracked files. Silent no-op class, on the Mac, in the script whose entire job is not clobbering things.

## Test plan

- [x] All 15 suites pass on `/bin/bash` 3.2.57 and `/bin/zsh` 5.9 / macOS 26.3.1: 434 assertions, 29s
- [x] Environment probes verified locally: `stat -c` fails, `date -d` fails, `${a[-1]}` yields empty on 3.2
- [x] Lint self-test: 32 passed, 0 failed, under bash 3.2 (macOS), bash 5.2 + gawk and bash 5.2 + mawk (Debian)
- [x] `shellcheck` clean on both new scripts at every severity, not just `error`
- [x] Both workflow files parse as YAML
- [x] Discrimination proven locally: bug 1 reintroduced is red on 3.2 and green on 5.2
- [x] `macos-bash32` runs and reports on this PR: 13 suites, 365 assertions, 44s ([log](https://github.com/scrollinondubs/behalfbot/actions/runs/33971656089/job/101321127544)). Re-run with the two zsh suites added is on the current head
- [x] Watchdog verified: a suite that sleeps forever is killed, reported as `HUNG` with RC 143, its partial output still printed, and no orphaned `sleep` left behind
- [x] `macos-bash32` goes red against a reintroduced bug 1 while ubuntu stays green, same commit ([red](https://github.com/scrollinondubs/behalfbot/actions/runs/33971770842/job/101321431211) / [green](https://github.com/scrollinondubs/behalfbot/actions/runs/33971770842/job/101321431075))

`chassis/VERSION` deliberately untouched.


jacketyjax:
## CI evidence

### The lane runs, and it runs something

Green run on the real commit: https://github.com/scrollinondubs/behalfbot/actions/runs/33971656089/job/101321127544

```
which -a bash:
/Users/runner/work/_temp/bash32/bash
/bin/bash
bash          -> 3.2.57(1)-release
env bash      -> 3.2.57(1)-release
/bin/bash     -> 3.2.57(1)-release

ok  chassis/scripts/test-chassis-update-health.sh (16 assertions)
ok  chassis/scripts/test-plugin-root-resolution.sh (44 assertions)
ok  chassis/scripts/test-chassis-root-health.sh (8 assertions)
ok  chassis/scripts/test-container-liveness.sh (7 assertions)
ok  chassis/scripts/test-chassis-root-resolution.sh (35 assertions)
ok  chassis/scripts/test-bootstrap-audit-memory.sh (12 assertions)
ok  chassis/scripts/test-memory-canary.sh (18 assertions)
ok  chassis/scripts/test-chassis-update-check.sh (46 assertions)
ok  chassis/scripts/test-chassis-update-mode.sh (54 assertions)
ok  chassis/scripts/test-discord-reply-param-fix.sh (24 assertions)
ok  chassis/scripts/test-colima-ensure.sh (31 assertions)
ok  chassis/scripts/test-control-flags.sh (30 assertions)
ok  plugins/dating/tests/test-emulator-state.sh (40 assertions)

suites run: 13   assertions: 365
all 13 suites pass on bash 3.2 + BSD userland
```

365 on the runner, the same 365 measured locally. Not a job that is green because it ran nothing.

One honest note on the pin: `which -a bash` on today's `macos-latest` image shows only the shim and `/bin/bash`, no Homebrew bash. So the shim was not load-bearing on this image. The assertion step is what keeps that true if the image changes, which is why it fails the job rather than warning.

`portability` on the same push: 30/30 self-test, then `12 finding(s) across 115 shell file(s)` - byte-identical to the local run.
https://github.com/scrollinondubs/behalfbot/actions/runs/33971656069/job/101321127193

### The lane discriminates, on CI, on one commit

Commit `9399658` put `fallback_root="${BAKED_ROOTS[-1]:-}"` back into `resolve-plugin-root.sh`. Both jobs below are from workflow run `33971770842`, the same commit, the same suite:

| Job | Result | |
|---|---|---|
| `plugin-root-resolution` (ubuntu, bash 5.x) | **pass**, 6s | [log](https://github.com/scrollinondubs/behalfbot/actions/runs/33971770842/job/101321431075) |
| `macos-bash32` (bash 3.2.57, BSD) | **fail**, 39s | [log](https://github.com/scrollinondubs/behalfbot/actions/runs/33971770842/job/101321431211) |

```
FAIL [s8 falls back to baked] expected '/var/folders/.../s8/baked', got ''
test-plugin-root-resolution: 43 passed, 1 failed
...
suites run: 13   assertions: 364
FAILED: chassis/scripts/test-plugin-root-resolution.sh
```

The empty plugin root reproduces, and exactly one of the two lanes sees it. Note the totals: 364, not 44. All thirteen suites still ran and reported after the failure, so one red suite does not hide the other twelve.

### Timing

| | |
|---|---|
| `macos-bash32` | **44s** |
| slowest ubuntu job in the same workflow (`dating-emulator-state`) | 15s |
| `Shell unit tests` workflow, this PR | 54s |
| `Shell unit tests` workflow, recent runs on main and other branches | 17s - 29s |
| `build` (docker-publish), this PR | **3m 05s** |

So the shell-tests workflow goes from roughly 25s to 54s, and **PR turnaround does not move at all**: `build` is the long pole at over three minutes and the macOS lane finishes inside the first quarter of it. Cost is zero either way - the repo is public, so the 10x macOS multiplier bills against free minutes.

### Filed, not fixed here

The 12 advisory findings are all pre-existing. Two are live defects with their own issue, the rest are grouped:

- #200 - `heartbeat-dispatcher.sh:389` parses `last_checked` with BSD-only `date -j`, so **every `every Nh` heartbeat matches on every tick** inside the container. Verified against the reference install's own dispatcher log, not inferred: `books-import` is `every 6h` and matched on all 45 ticks of 2026-09-05.
- #201 - `migrate-from-old-clone.sh:146` uses `mapfile` on a host-side script. On bash 3.2 the array is empty, the dynamic exclude list is empty, and the migration clobbers chassis-tracked files.
- #202 - the remaining eight: `guardrails.sh:131` (bug 2's shape), four BSD-date sites in `welfare-check-gather.sh`, an unguarded `sha256sum`, `declare -A` in `hydrate-env-from-vw.sh`, and a bare `timeout` in the dispatcher.

### One wart

Force-push is blocked in my environment, so commit `9399658` (the deliberately broken demo) could not be rewritten out - it is undone by a revert instead. The tree at HEAD is byte-identical to the single good commit, and `chassis/scripts/resolve-plugin-root.sh` has a zero-line diff against `origin/main`. **Squash-merge** and the history stays clean.

jacketyjax:
## Final CI, and the failure this lane found

### Green on the current head

[`macos-bash32`, run 33972981942](https://github.com/scrollinondubs/behalfbot/actions/runs/33972981942/job/101324670776) - **success, 55s**:

```
bash          -> 3.2.57(1)-release
env bash      -> 3.2.57(1)-release
/bin/bash     -> 3.2.57(1)-release
...
suites run: 15   assertions: 434
all 15 suites pass on macOS: bash 3.2.57, zsh 5.9, BSD userland
```

`Shell unit tests` completed/success. `Shell scripts lint` completed/success, with the `portability` self-test at 32/32 and 12 advisory findings.

### The lane's first real finding

Adding the three zsh `heartbeat-fire-caps` suites from #196 for parity made `test_fire_caps_e2e` run for **8+ minutes** on a macOS runner against 22 seconds on ubuntu ([run 33972207796](https://github.com/scrollinondubs/behalfbot/actions/runs/33972207796), 14:34:24 to a manual cancel at 14:42:45).

Cause, reproduced locally with nothing but `PATH` changed - 42s with Homebrew present, hung and killed at 100s under `PATH=/usr/bin:/bin:/usr/sbin:/sbin`:

`heartbeat-dispatcher.sh:77` resolves its watchdog as `command -v gtimeout || command -v timeout || echo timeout`. Both are GNU coreutils and neither exists on a stock macOS install, so `TIMEOUT_CMD` becomes the literal string `timeout`, `$TIMEOUT_CMD 1200 /bin/zsh -c` around the `claude -p` call returns 127, and the retry policy sleeps 20s then 60s per attempt. `ps` mid-run:

```
26989 26974  zsh .../heartbeat-dispatcher.sh
27125 26989  sleep 20
```

**On a fresh Mac install no heartbeat can invoke a model at all.** Invisible until now because the reference install has Homebrew on PATH. Filed as **#203** with the reproduction, not fixed here. That suite is the single macOS-only exclusion, with the issue cited inline and an explicit note not to re-enable it by installing coreutils on the runner, which would remove the userland the lane exists to exercise.

Two things this PR gained as a result, and it should have had both from the start:

- **180-second per-suite watchdog**, hand-rolled, because GNU `timeout` is exactly the binary this platform does not have. Verified: a suite that sleeps forever is killed, reported as `HUNG` with RC 143, its partial output still printed (the previous shape only `cat`'d the log after the suite exited, so a hang produced nothing at all), and no orphaned `sleep` left behind.
- **`timeout-minutes: 15`** on the job as a backstop.

Finding a hang now costs three minutes rather than a six-hour runner.

### Timing, final

| | |
|---|---|
| `macos-bash32` | **55s** (15 suites, 434 assertions) |
| slowest ubuntu job in the same workflow (`plugin-root-resolution`) | 44s |
| `Shell unit tests` workflow, this PR | 1m 07s |
| same workflow, recent runs on main | 17s - 29s |
| `build` (docker-publish), the long pole | 3m+ |

PR turnaround does not move: the macOS lane finishes inside the first third of `build`.

### Issues opened

- **#200** - dispatcher parses `last_checked` with BSD-only `date -j`, so every `every Nh` heartbeat matches on every tick in the container. Verified against the reference install's dispatcher log: `books-import` is `every 6h` and matched on all 45 ticks of 2026-09-05.
- **#201** - `migrate-from-old-clone.sh:146` uses `mapfile` on a host-side script; on bash 3.2 the exclude list is empty and the migration clobbers chassis-tracked files.
- **#202** - the remaining eight advisory findings.
- **#203** - the dispatcher cannot invoke Claude at all on a stock macOS host. Found by this lane.

Ready for review. Not merging - **squash**, so the cancelled demo commit `9399658` does not land in history.

